PR #62, opened 2014-01-20
Allow PHPUnit tests to be run from any directory (relative path)

scottchiefbaker:
This makes the Test.php callable from any path, not just the **parent** parsedown directory.


hkdobrev:
I suggest this to be closed and if the PR is still needed a new one could be opened. This PR is from `master` branch which is now updated with `erusev:master` and does not have any files changed. A possible new PR could be from a feature branch which would make more sense.

PR #76, opened 2014-01-30
Add a CLI mode to allow parsing markdown at the command line

scottchiefbaker:
I was writing a lot of tests, and then copy/pasting the HTML from my browser to the appropriate HTML file, which was very time consuming. Then it occurred to me that we could use Parsedown.php **as** the CLI tool. With this code you can now do:

```
php Parsedown.php document.md > document.html
```

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -13,6 +13,14 @@
 #
 #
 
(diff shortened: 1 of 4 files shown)


aidantwoods:
It seems this was superseded by #275

PR #79, opened 2014-01-30
Add .editorconfig for cross-editor configuration

hkdobrev:
See http://editorconfig.org/ for more.

Also added newline characters at the end of the files.

---

EditorConfig is supported on a number of text editors and IDEs. Including, but not limited to:
- Sublime Text
- Vim
- Emacs
- TextMate
- JetBrains IDEs
- Visual Studio IDEs


.editorconfig
```diff
@@ -0,0 +1,11 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
```
aidantwoods: Based on what we do in the tests I think we should do something like:

```config
root = true

[*]
charset = utf-8
end_of_line = lf
insert_final_newline = true
indent_style = space
indent_size = 4
trim_trailing_whitespace = true

[*.md]
trim_trailing_whitespace = false
insert_final_newline = false

[*.html]
trim_trailing_whitespace = false
insert_final_newline = false
```

hkdobrev:
I'll try to update my open PRs next week.

aidantwoods:
Awesome! Just whenever you have the time – no pressure :)

PhrozenByte:
IDE config files do not belong in public repositories, no matter whether we're talking about config files of a single IDE, or some "pseudo"-standard config files.

aidantwoods:
@PhrozenByte In general I completely agree that IDE config files do not belong in public repositories.

However, I personally think that `.editorconfig` deserves to be an exception to this rule:

The primary function of `.editorconfig` is to communicate code-styling guidelines. Whatever editor you use—even something like notepad—the file serves this purpose as a human readable set of rules that should be followed for the repo. This is just one of many ways of communicating these guidelines (depending on how much you care about their enforcement: other methods include failing CI builds if style guidelines aren't met, or being caught in manual code review).

In my mind the fact that some editors support reading this file and changing their configuration is a convenience more than anything else,`.editorconfig` definitely isn't designed to be a primary configuration file. The file still serves a good purpose without editor support for it, because it still serves the role of documenting code-style.

If someone really doesn't like their editor settings changing when navigating to other's folders, `.editorconfig` probably isn't something that they'd ever want, and hopefully if reading this file is undesired, an editor can be instructed to ignore it.

PhrozenByte:
> The primary function of `.editorconfig` is to communicate code-styling guidelines.

The problem is: That's not what it does. It just tells IDEs to use Unix line feeds and 4 space intendation - something virtually *any* IDE already ascertains by simply looking at the file. This solves a non-existent problem.

PR #86, opened 2014-02-03
restrict the maximum recursion depth

wkpark:
it might be helpful to restrict the maximum recursion depth to protect recursion attacks with nested markup like as following

```
> > > > > > ..... > > infinite block quotes
or
* * * * * ... * * infinite list

> * > * > * > * > * ... * nested blockquote + list
```

> > > > > > > > > > > blah blah


aidantwoods:
This seems like a good idea, unfortunately it's become really out of date. Happy to look at an updated version though :)

PR #95, opened 2014-02-06
Allow a single space in the code span delimiter

### LINKED ISSUE #89: Support spaces as delimiters in inline code spans
From the [Markdown spec](http://daringfireball.net/projects/markdown/syntax#code):

> The backtick delimiters surrounding a code span may include spaces — one after the opening, one before the closing. This allows you to place literal backtick characters at the beginning or end of a code span.

Here is a failing test:

```
A single backtick in a code span: `` ` ``

A backtick-delimited string in a code span: `` `foo` ``
```

``` html
<p>A single backtick in a code span: <code>`</code></p>
<p>A backtick-delimited string in a code span: <code>`foo`</code></p>
```

---

Parsedown includes the spaces inside the `<code>` element. These spaces would eventually be rendered literally by a browser.

``` html
<p>A single backtick in a code span: <code> ` </code></p>
<p>A backtick-delimited string in a code span: <code> `foo` </code></p>
```

hkdobrev:
Fix #89 properly.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1028,9 +1028,9 @@ class Parsedown
 
                 case '`':
 
-                    if (preg_match('/^(`+)[ ]*(.+?)[ ]*\1(?!`)/', $text, $matches))
+                    if (preg_match('/^(`+)([ ]?)(.+?)\2\1(?!`)/', $text, $matches))
                     {
-                        $element_text = $matches[2];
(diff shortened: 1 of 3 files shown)


aidantwoods:
I tried the given test cases in Parsedown and they agree with the CommonMark reference parser (but not the given test case results here). I guess these were based on the Gruber spec? ;p

PR #108, opened 2014-02-16
Fix #107 : em and strong containing html link with _ inside

### LINKED ISSUE #107: em and strong fail if html with _ inside
This test case works fine
`___em strong <a href="http://parsedown.org/" class='externallink'>Parsedown</a>___`

Whereas this fails
`___em strong <a href="http://parsedown.org/" class='external_link'>Parsedown</a>___`

Cerdic:


aidantwoods:
As mentioned in https://github.com/erusev/parsedown/issues/588#issuecomment-377221179, we can't really fix the core issue properly in a reasonable way. We could fix this one case, but it starts to become a bit of a slippery slope towards putting awareness about all inline elements and how they start and end into just this one element so that it knows how to skip over them (as mentioned in https://github.com/erusev/parsedown/issues/107#issuecomment-377276537).

Personally I think that in this instance, if we're not going to be compliant with the spec then it is better to be predictably non-compliant than to have a selection of special cases that work because we thought of them.

I hope you understand. Thank you nonetheless for taking the time to diagnose and write a patch here :)

PR #177, opened 2014-05-28
Added safe mode

clphillips:
I've added new unit tests as well, for which this patch does not pass completely due to the expected unsafe output in the existing markdown output results. These tests should probably be ignored.

As the unit tests show, there's also an issue with plain email conversion due to the < and > around the email address. Still a work in progress, but I figured best commit what I have since others are working on the issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+parsedown.komodoproject
+test.php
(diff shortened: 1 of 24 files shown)


aidantwoods:
#495 added safe-mode in the AST :)

PR #194, opened 2014-07-11
Load markdown text from a file.

AlienEngineer:
I'm not very php oriented but I believe this is not that wrong. I used this to parse a bunch of markdown files into an Html page.


aidantwoods:
I think this is maybe best done by the user :)

You can always write your own function to instantiate, configure, and call Parsedown's methods on your behalf if you find yourself repeating the same lines of code a lot.

PR #205, opened 2014-08-22
fix utf8 string handling

kirbach:
substr_replace() is not multibyte safe. Having utf8 encoded characters in the markup, breaks the html output.


aidantwoods:
Hi, thanks for the PR!

There was some similar work done in #513, so I'm going to close this in favour of that since it aims to bring multibyte support to all the structures in Parsedown (I think it's probably about time we patch this everywhere). I hope this makes sense :)

PR #212, opened 2014-09-02
Automatically encode all email links using as hex HTML entities

### LINKED ISSUE #193: Encoded emails when using 'autolink'
It would be great if you could add support for encoding emails as specified in Gruber's spec.

http://daringfireball.net/projects/markdown/syntax#autolink

Currently, `<address@example.com>` gets parsed semi-correctly to `<a href="mailto:address@example.com">address@example.com</a>`

It should be parsed into this, though:

```
<a href="&#x6D;&#x61;i&#x6C;&#x74;&#x6F;:&#x61;&#x64;&#x64;&#x72;&#x65;
&#115;&#115;&#64;&#101;&#120;&#x61;&#109;&#x70;&#x6C;e&#x2E;&#99;&#111;
&#109;">&#x61;&#x64;&#x64;&#x72;&#x65;&#115;&#115;&#64;&#101;&#120;&#x61;
&#109;&#x70;&#x6C;e&#x2E;&#99;&#111;&#109;</a>
```

You can see the difference on your [demo page](http://parsedown.org/demo).

Thanks and keep up the great work!

apfelbox:
It currently only encodes the `href`, not any content of the node.

So something like

```
[mail@example.org](mailto:mail@example.org) will only encode the link.
The same goes for <mail@example.org>.
```

Although in the latter case one could actually argue to also encode the text.
Thoughts?

Fixes #193

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1135,7 +1135,7 @@ class Parsedown
                     'name' => 'a',
                     'text' => $matches[1],
                     'attributes' => array(
-                        'href' => 'mailto:'.$matches[1],
+                        'href' => $this->encodeEmailUrl('mailto:' . $matches[1]),
                     ),
                 ),
             );
@@ -1230,6 +1230,11 @@ class Parsedown
         }
 
         $url = str_replace(array('&', '<'), array('&amp;', '&lt;'), $Link['url']);
+
+        if ("mailto:" === substr($url, 0, 7))
+        {
+            $url = $this->encodeEmailUrl($url);
+        }
 
         if ($Excerpt['text'][0] === '!')
         {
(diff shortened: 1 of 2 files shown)


aidantwoods:
I think this is out of scope for now since it isn't part of CommonMark or the GitHub Flavoured extension.

Just to comment about email obfuscation in general: IMO it is better to display the email in plaintext without obfuscation so that users can see that the email is given out in plaintext. It's trivial to write a decoder for email addresses obfuscated in this way, and I think the worst thing of all would be for a user to believe that they somehow had not given their email address out publicly because of obfuscation.

If you give out your email address, encoded or not, you've still given out your email address. I think we shouldn't hide this fact.

The appearance of safety or security where there is none is likely to be worse than none on its own, because it will likely encourage the user to behave as if there were some in place.

PR #237, opened 2014-10-10
Xhtml disabler

naNuke:
Switch between xhtml tag closers `/>` to `>`

This is pretty much open to discussion, I use it in my project and it works perfect so far :)


laukstein:
@aidantwoods please add this or best replace closer tag `/>` with `>`. Know that XHTML isn't no longer supported in HTML5 by spec.

PR #249, opened 2014-12-10
Updated readme to display travis-ci.org build test

HelpingHand1:


hkdobrev:
:+1: :tada:

Recently I've read an article asking to keep the build green with a very good point. Imagine you always run the tests before tagging a new release. But your `master` branch should have a green build as well, because otherwise you would prevent collaborators to easily make contributions.

So if the tests on the `master` branch fail and someone make a fork, make a change and wants to submit a PR, the tests would fail for them too. And they could even spend some time investigating what they broke until they realize the tests were failing even without their changes.

Having a build badge in the readme shows the maintainers and the possible collaborators if anything is broken on master so either they fix it or at least to keep it in mind.

If however there is some larger changes you'd like to make which are going to break the build it is best to have them in another branch, because the `master` would be used from collaborators to base their work.

erusev:
@hkdobrev Good points. Thanks for the input.

PR #258, opened 2015-01-10
Added missing require block

GrahamCampbell:
NB, composer doesn't even run on php 5.2 so there'd be no point in allowing it in the version constraint.

Also, I think it's irresponsible to support php 5.2 given it's oel and there are known vulnerabilities

---
http://blog.ircmaxell.com/2014/12/on-php-version-requirements.html
http://blog.ircmaxell.com/2014/12/php-install-statistics.html

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,6 +12,10 @@
             "homepage": "http://erusev.com"
         }
     ],
+    "require": {
+        "php": ">=5.3.3",
+        "ext-mbstring": "*"
+    },
     "autoload": {
         "psr-0": {"Parsedown": ""}
     }


GrahamCampbell:
Ping.

erusev:
Thanks for the PR. I haven't had a chance to look at the links. I'll do so as soon as I resolve the issues that I've started working on.

GrahamCampbell:
Ping @erusev.

erusev:
> Also, I think it's irresponsible to support php 5.2 given it's oel and there are known vulnerabilities.

When confronted with the fact that Parsedown doesn't support their version of php, some people would perhaps upgrade, but others might not. The could instead consider using another parser instead or even another markup language (e.g. bbcode). The restrictions might promote newer versions of php, but along with that it might do the opposite for the adoption of Markdown. I do understand the importance of upgrading php and I might consider adding a footnote that encourages that, but I'm not completely convinced that forcing people to upgrade is a solution that I'd feel comfortable with.

GrahamCampbell:
People using php 5.2 should be totally ashamed.

hkdobrev:
@erusev I agree with your arguments, I just want to add that this PR would not stop anyone from using it on PHP 5.2.

GrahamCampbell:
Can we get this merged then? Composer itself requires php 5.3.2, so I don't see what the issue is?

cebe:
specifying the php version in a composer.json does not really make sense if parsedowns version requirement is lower than composers...

GrahamCampbell:
Why would that be. It would be incorrect for us not to include a version constraint on php.

cebe:
then the correct one would be `>=5.2`

PR #260, opened 2015-01-11
Fix for Parsedown stripping classes on images supported by ParsedownExtra

rhukster:
More extensive detail of what's going on here:

https://github.com/erusev/parsedown-extra/issues/32

However, in short: With the refactoring in Parsedown 1.2.0, `inlineImage()` and `inlineLInk()` were split out into two separate methods.  This for the most part works great, but with ParsedownExtra's class support for images, it becomes a problem.  The class attributes are stripped out in Parsedown's `inlineImage()` method because it creates a non-markdown-extra formatted Inline element from scratch.

My solution, although there may be a better one, is to manipulate the Inline element so the class attribute is not lost (if it exists).

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1298,6 +1298,8 @@ class Parsedown
 
         $Inline = $this->inlineLink($excerpt);
 
+        $backup = $Inline;
+
         if ($Inline === null)
         {
             return;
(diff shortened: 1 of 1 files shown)


rhukster:
Travic CI failures because the order of the tags is not maintained with this approach:

```
1) ParsedownTest::test_ with data set #31 ('image_title', '/home/travis/build/erusev/par.../data/')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'<p><img src="/md.png" alt="alt" title="title" /></p>'
+'<p><img title="title" src="/md.png" alt="alt" /></p>'
```

erusev:
I'll deal with the failing tests.

rhukster:
Great thanks!

PR #275, opened 2015-01-20
Add a CLI

hkdobrev:
- Super simple CLI with the ability to read from files and STDIN.
- Add [vendor binary to
`composer.json`](https://getcomposer.org/doc/articles/vendor-binaries.md).
- Add CLI usage to README.

Closes #76.


parsedown
```diff
@@ -0,0 +1,24 @@
...
+require(dirname(__FILE__).'/Parsedown.php');
+
+// From a file
+if (isset($argv[1])) {
+
+	if ( ! is_readable($argv[1])) {
+		throw new RuntimeException(sprintf(
+			'Could not read "%s"',
+			$argv[1]
+		));
```
cebe: how does the exception appear to the user of the script? I would prefer a more userfriendly echo + exit(1) here.

hkdobrev: I thought of that, but I'd like users to see what's the actual exception. And I thought throwing an exception in a PHP CLI would output it and exit with non-zero status code. But I need to verify that actually. Perhaps an option should be enabled. I'll think about that, thanks!

cebe:
```php
<?php
throw new RuntimeException('Could not read "..."');
```
renders as:
```
$ php php.php 
PHP Fatal error:  Uncaught exception 'RuntimeException' with message 'Could not read "..."' in /tmp/php.php:3
Stack trace:
#0 {main}
  thrown in /tmp/php.php on line 3
```

which could look much nicer, in this case it does not matter what kind of exception it was and in which line of the script itwas thrown.

hkdobrev: @cebe I've used `echo` and `exit`.

hkdobrev:
@erusev @cebe @scottchiefbaker I've updated the PR a bit. I am not sure if the `parsedown` binary is better off in a `bin/` folder or in the root of the repo. What do you think?

parsedown
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env php
+<?php
+
+require(dirname(__FILE__).'/Parsedown.php');
+
+// From a file
+if (isset($argv[1])) {
+
+	if ( ! is_file($argv[1])) {
+		printf('"%s" is not a file', $argv[1]);
```
cebe: Both printf should add a newline at the end.

cebe:
I do not think a bin subdirectory is needed.

parsedown
hkdobrev: On PHP 5.6 it adds a newline by itself. Do you mean you want another one?

cebe: not for me and not on other systems according to 3v4l... http://3v4l.org/5pvgL

scottchiefbaker:
I've got a tool written on my profile: [MarkdownCLI](https://github.com/scottchiefbaker/MarkdownCLI)

Issue #210, opened 2014-08-28
XSS Image Tag


xLink:
Hey im using your parser in a laravel install, just had someone post 
```md
![" onerror="alert(1)](0 ""alert(2);")
``` 

and it return
 
```html
<img alt="" onerror="alert(1)" src="0" title=""alert(2);" />
```
with this the alert(1) was fired, is there anyway to have this fixed? I am already doing an escape within my code before the markdown gets parsed, but ofc it wouldnt pick this up :/

erusev:
I'm closing this as a duplicate of #161.

Issue #161, opened 2014-04-29
Safe Mode


clphillips:
Parsedown offers no safe option for rendering, so, for those of us who allow **no** HTML content within markdown text, we must sanitize prior to feeding to Parsedown. As a result, Parsedown double encodes HTML entities (see #50).

Instead, Parsedown should offer a "safe" option.

``` php
function text($text, $safe_mode = false) {
    $this->safe_mode = $safe_mode;
    if ($safe_mode) {
        $text = htmlentities($text, ENT_QUOTES, 'UTF-8');
    }
    ...
}
```

Usage:

``` php
$parsedown = new ParseDown();
$text = "<script>alert('unsafe text');</script>";
echo $parsedown->text($text, true);
```

Output:

```
&lt;script&gt;alert(&quote;unsafe text&quote;);&lt/script&gt;
```

ckzhou:
[xss](http://"onmouseover="alert(1))
This vector can pop a alert box in firefox!

xorock:
For several years I used Textile https://github.com/textile/php-textile and there is textileRestricted method which is suitable for untrusted input. As I understand there is no such thing in this library?

erusev:
Would someone like to create a simple set of test data that includes `.md` and `.html` files with input and expected safe output.

clphillips:
@erusev You want that as a PR, or what? Examples here: [XSS Filter Evasion Cheat Sheet](https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet).

erusev:
@clphillips Sure, but let's start simple, 5 to 10 pairs to see if this test driven approach would work.

erusev:
Perhaps, we could start with something like: https://github.com/markdown-it/markdown-it/blob/019bbda5f5ee8b7d00f2633340aef3b0d000e3f1/test/fixtures/markdown-it/xss.txt

dariuszp:
Hi, any ETA on this? I need markdown in my latest project. I was thinking of using parsedown but it will be for user generated content.
Because of this issue I had to abandon this library.

BboyKeen:
I'm interested too. Any news for the safe mode ?

erusev:
I've been thinking about this and I do plan to implement it, hopefully in January.

hootlex:
any update on this?

erusev:
Not yet, but perhaps soon.

hootlex:
Great, looking forward.

[...]

PR #276, opened 2015-01-21
Prevent various XSS attacks

naNuke:
Hello, I am back with the safe links issue. 
As always this is open to discussion :) but together with already existing markupEscaped setting, this is basically a fix for issues #161 and #210, and maybe other im not aware of.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -84,6 +84,14 @@ class Parsedown
 
     protected $safeLinksEnabled = true;
 
+    protected $safeLinksWhitelist = array(
+        'http://',
+        'https://',
+        '/',
+        'ftp://',
+        'ftps://'
+    );
+
     #
     # Lines
     #
@@ -1244,9 +1252,22 @@ class Parsedown
             $Element['attributes']['title'] = $Definition['title'];
         }
 
-        if ( $this->safeLinksEnabled && preg_match("/^(\/|https?:\/\/|ftps?:\/\/)/ui", $Element['attributes']['href']) === 0 )
+        if ( $this->safeLinksEnabled )
         {
-            return;
+            $matched = false;
+            foreach ( $this->safeLinksWhitelist as $scheme )
+            {
+                if ( stripos($Element['attributes']['href'], $scheme) === 0 )
+                {
+                    $matched = true;
+                    break;
+                }
+            }
+
+            if ( ! $matched )
+            {
+                return;
+            }
         }
 
         $Element['attributes']['href'] = htmlspecialchars($Element['attributes']['href'], ENT_QUOTES);
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1244,13 +1244,18 @@ class Parsedown
             $Element['attributes']['title'] = $Definition['title'];
         }
 
-        if ( $this->safeLinksEnabled && stripos($Element['attributes']['href'], 'javascript:') === 0 )
+        if ( $this->safeLinksEnabled && preg_match("/^(\/|https?:\/\/|ftps?:\/\/)/ui", $Element['attributes']['href']) === 0 )
         {
             return;
         }
 
-        $Element['attributes']['href'] = htmlspecialchars($Element['attributes']['href']);
-        $Element['text'] = htmlspecialchars($Element['text']);
+        $Element['attributes']['href'] = htmlspecialchars($Element['attributes']['href'], ENT_QUOTES);
+        $Element['text'] = htmlspecialchars($Element['text'], ENT_QUOTES);
+
+        if ( $Element['attributes']['title'] !== null )
+        {
+            $Element['attributes']['title'] = htmlspecialchars($Element['attributes']['title'], ENT_QUOTES);
+        }
 
         return array(
             'extent' => $extent,
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -75,6 +75,15 @@ class Parsedown
 
     protected $urlsLinked = true;
 
+    function setSafeLinksEnabled($safeLinksEnabled)
+    {
+        $this->safeLinksEnabled = $safeLinksEnabled;
+
+        return $this;
+    }
+
+    protected $safeLinksEnabled = true;
+
     #
     # Lines
     #
@@ -1235,7 +1244,13 @@ class Parsedown
             $Element['attributes']['title'] = $Definition['title'];
         }
 
-        $Element['attributes']['href'] = str_replace(array('&', '<'), array('&amp;', '&lt;'), $Element['attributes']['href']);
+        if ( $this->safeLinksEnabled && stripos($Element['attributes']['href'], 'javascript:') === 0 )
+        {
+            return;
+        }
+
+        $Element['attributes']['href'] = htmlspecialchars($Element['attributes']['href']);
+        $Element['text'] = htmlspecialchars($Element['text']);
 
         return array(
             'extent' => $extent,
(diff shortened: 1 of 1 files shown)


Parsedown.php
```diff
@@ -75,6 +75,15 @@ function setUrlsLinked($urlsLinked)
 
     protected $urlsLinked = true;
 
+    function setSafeLinksEnabled($safeLinksEnabled)
+    {
+        $this->safeLinksEnabled = $safeLinksEnabled;
```
hkdobrev: You'd want to declare this property as a `protected` one and not just leave it to be declared in runtime.

hkdobrev: Sorry, saw it below later.

erusev:
This looks nice. I'd love to hear more opinions on this. I don't feel like an expert on XSS.

xPaw:
I feel like this param should be named as something more generic like `safeAttributes` and escaping every attribute that user can enter (e.g `title`). Also I think checking for `javascript:` in urls seems like a silly idea, and urls should be escaped regardless.

erusev:
@xPaw Thanks for the input. Which characters do u think should be escaped?

xPaw:
Using `htmlspecialchars` should do the job just fine.

erusev:
Thanks! 

> escaping every attribute that user can enter (e.g title)

Is this about attributes of anchor (`<a>`) tags or all tags?

xPaw:
Any attribute that user can change should be escaped. I guess in Parsedown that only applies to links really.

erusev:
What if instead of having an option about it we implementing it as a default behaviour. It shouldn't have too much performance cost.

xPaw:
That actually would make more sense.

naNuke:
We dont need to escape title attribute because of the way how parsedown works, any additional quote simply breaks the regex and link wont render at all.

thing like ```[example](www.example.com "><script>alert('xss')</script>")```
ends up as ```<a href="www.example.com" title="><script>alert('xss')</script>">```

But adding escape on title is easy if there is really need for it.

The check against javascript: in url is to prevent people from writing clickable javascript links which can be used as an XSS attack.

erusev:
@xPaw Would processing attributes through `htmlspecialchars` protect against things like `[link](javascript:alert('xss'))`?

xPaw:
It won't, but simply checking if the string starts with `javascript:` won't fix the problem either.

See http://security.stackexchange.com/a/73751 for some examples as to why.

erusev:
I asked SO for advice at http://stackoverflow.com/q/28129788/200145

naNuke:
Updated with info from @xPaw to whitelist the protocols instead of javascript: check.
Didn't know the pseudo-protocol can be malformed in some browsers ^^
Also escaping title just to be 100% sure.

xPaw:
I don't think having hardcoded schema list in the regex is a way to go. For example reddit has a bigger schema whitelist: https://github.com/reddit/reddit/blob/bb0f5b552c3a52abfac1e2147144eef8f2e7dfab/r2/r2/lib/filters.py#L177

naNuke:
Thanks for input :) added customizable whitelist, but since parsedown is parser meant for everyone I am not sure if it should include all possible schemas by default. Thoughts?

hkdobrev:
Here a few things I think should be done in order Parsedown to be more safe against XSS.

#### HTML

Raw HTML should be disabled in Markdown. You could use `setMarkupEscaped()` in Parsedown.

#### User-generated content

If a developer is getting user-generated content and is putting it inside Markdown documents this should be escaped/validated beforehand. You could also use HTML Purifier or a similar library to actually validate/escape content after the HTML is generated.

#### Attributes

Every attribute should be escaped for special characters, not only links. In Parsedown this would be the `href`, `title`, `src` and `alt` attributes.

#### Schemas:

1. There is no definitive list of schemas, because you can create your own.
2. [According to OWASP](https://www.owasp.org/index.php/XSS_(Cross_Site_Scripting)_Prevention_Cheat_Sheet#XSS_Prevention_Rules_Summary) you should whitelist `http://` and `https://` only for security purposes.

---

Don't take this the wrong way, I think this is a step in the right direction.

About what @erusev suggested that this might be the default:

- Attributes escaping could be done by default.
- HTML escaping is already behind an option and of course should stay that way.
- Schema whitelisting definitely should be optional.
- A note on which says Parsedown is not enough for your content to be secure would be good. Users which use Parsedown should be educated to think about their content and which text comes from where.

naNuke:
Is there any other markdown syntax that accepts user content directly for its generated html other than links and images?

naNuke:
I've just finished reading thru the code and added escaping on *hopefully* all parsed attributes.

On another note, at the moment, all of this expects user input to be UTF-8, which may leave some project vulnerable by not catching html entities under different encodings. [Read here for more info](http://blog.astrumfutura.com/2012/03/a-hitchhikers-guide-to-cross-site-scripting-xss-in-php-part-1-how-not-to-use-htmlspecialchars-for-output-escaping/)

I am not quite sure how to solve the encoding issue since that might be out of scope of this fix, also the fact that parsedown is supporting php 5.2 is not helping, but my idea of possible solution is:

- text() method would accept encoding parameter, encode from that encoding to utf-8, go through all the lines, then re-encode back to user encoding and return.

Not aware of any complications this might bring, but I am not that knowledgeable about unicode.

Let me know what you guys think :)

hkdobrev:
:star2:  LGTM :sparkles: :dizzy:

hkdobrev:
@erusev What's up with this?

erusev:
@hkdobrev I have some thoughts about it. Hopefully, I'll have more time next week to write about them.

ivantcholakov:
Maybe borrowing code from CodeIgniter 3 would be the solution. PHP min = 5.2.4
https://github.com/bcit-ci/CodeIgniter/blob/3.0.0/system/core/Security.php#L349

adamthehutt:
Is there any update on this? Anything I can do to help?

hkdobrev:
> Is there any update on this? Anything I can do to help?

:+1:

erusev:
@adamthehutt You could help with creating simple test data - .md / .html pairs that contain XSS input and the expected safe output.

ifeltsweet:
LGTM

malas:
why this is still not merged into?

erusev:
@malas one reason is the lack of tests

malas:
@erusev then is it safe to say that his project is not supported? such a major issue is not addressed for almost a year, does not sound like a ready for production project.

erusev:
@malas Parsedown is a Markdown parser, its responsibility is to parse Markdown, implementing a safe mode would be nice but it has nothing to do with wether the parser is supported or not so no, it would not be safe to say that the project is not supported

tillkruss:
Any updates on this?

aidantwoods:
If I may, I might suggest that encoding output should be the responsibility of the `element` method so that all attributes and content are always escaped in a consistent way (without having to hunt down all instances of where attributes are set, now or in future).

Choosing a safe URL (i.e. protocol whitelisting) should remain the responsibility of the specific link method though.

However, it may be a better approach to delegate this filtering process elsewhere, so that `href` in links and `src` in images are filtered to be safe at a higher level (e.g. so that extensions that completely override the link method still benefit from this). This argument would yield similar benefits on moving encoding to the `element` method.

aidantwoods:
Hi again everyone, I've gone ahead and implemented my own suggestion on top of the work started here (would be great to get Parsedown a safe mode – given how popular it is).

I've rebased this on master and opened a new PR (can't target this one without running into the conflicts I already had to resolve manually).

Please throw everything you've got at https://github.com/erusev/parsedown/pull/495, by trying to get some XSS through. Or come up with some more test cases if you like.

PR #284, opened 2015-02-03
Fixes #283

### LINKED ISSUE #283: Blank quoted title text in links and inline images doesn't parse properly
Some Markdown editors always insert a title for links and inline images, and just default it to a blank string when the user didn't input anything in the "Title:" prompt.  In a perfect world, these editors would just leave off the title argument when content wasn't provided, but this isn't something we can always depend on in [our app](http://www.cerbweb.com).

In previous versions of Parsedown, these blank titles parsed properly.  In the latest version, the regular expression requires content between the quotes or it doesn't parse the entire link.

The following properly converts to a link:

~~~
[link to google](http://www.google.com "This is the link title")
~~~

While this doesn't:

~~~
[link to google](http://www.google.com "")
~~~

This also breaks with inline images that have blank titles:

~~~
![blank title](/md.png "")
~~~

I've made the minor tweak and added new PHPUnit tests.  I'll submit the pull request after submitting this issue.

jstanden:
This commit fixes an issue where blank titles in links and inline images result in the link not parsing from Markdown into HTML at all.  As mentioned on #283, some Markdown editors always insert a title even when it's blank.  In [our project](http://www.cerbweb.com/), we have thousands of knowledgebase articles with these blank titles in the Markdown, which previously worked with PHP Markdown and slightly earlier versions of Parsedown.

The issue stems from the regular expression for links that looks for "one or more characters that aren't a quote" instead of "zero or more characters that aren't quotes".

Similarly, it would be perfectly valid in HTML to have ```<a href="http://www.example.com" title=""></a>```, even though it's pointless.

I added additional tests for blank link titles using single and double quotes, and inline images using blank titles.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1197,7 +1197,7 @@ class Parsedown
             return;
         }
 
-        if (preg_match('/^[(]((?:[^ ()]|[(][^ )]+[)])+)(?:[ ]+("[^"]+"|\'[^\']+\'))?[)]/', $remainder, $matches))
+        if (preg_match('/^[(]((?:[^ ()]|[(][^ )]+[)])+)(?:[ ]+("[^"]*"|\'[^\']*\'))?[)]/', $remainder, $matches))
         {
             $Element['attributes']['href'] = $matches[1];
 
--- a/test/data/image_title.html
+++ b/test/data/image_title.html
@@ -1,1 +1,2 @@
 <p><img src="/md.png" alt="alt" title="title" /></p>
+<p><img src="/md.png" alt="blank title" title="" /></p>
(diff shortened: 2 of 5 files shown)


hkdobrev:
@jstanden I would suggest to check if the title is not empty before adding the title attribute. E.g `if (!empty($matches[2])) ...`.

erusev:
@jstanden 

Thats a great pull request. Thanks!

@hkdobrev 

I'm in favour of that.

jstanden:
@erusev Sure thing.  Thanks for the great library.

@hkdobrev Yeah, that would be a good next step to generate cleaner HTML output.  :)  The regular expression had to be fixed to enter that block of code at all.

PR #285, opened 2015-02-08
Add "id" attribute to headings

henriquemoody:
This is more or less how GitHub Flavored Markdown works.

See an example: https://github.com/henriquemoody/parsedown/blob/titles/test/data/atx_heading.md

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -19,6 +19,8 @@ class Parsedown
 
     const version = '1.5.1';
 
+    private $headingSlugs = array();
+
     # ~
 
     function text($text)
@@ -462,6 +464,33 @@ class Parsedown
         return $Block;
     }
 
+    protected function slugify($text)
+    {
+        $slug = trim($text);
+        $slug = strtr($slug, ' ', '-');
+        $slug = strtolower($slug);
+
+        return preg_replace('/[^a-z0-9_-]/', '', $slug);
+    }
+
+    protected function getHeadingId($text)
+    {
+        $slug           = $this->slugify($text);
+        $attributeId    = $slug;
+
+        if (!isset($this->headingSlugs[$slug])) {
+            $this->headingSlugs[$slug] = 0;
+        }
+
+        if ($this->headingSlugs[$slug] > 0) {
+            $attributeId .= '-'.$this->headingSlugs[$slug];
+        }
+
+        $this->headingSlugs[$slug]++;
+
+        return $attributeId;
+    }
+
     #
     # Header
 
@@ -488,6 +517,9 @@ class Parsedown
                     'name' => 'h' . min(6, $level),
                     'text' => $text,
                     'handler' => 'line',
+                    'attributes' => array(
+                        'id' => $this->getHeadingId($text),
+                    ),
                 ),
             );
 
--- a/test/data/atx_heading.html
+++ b/test/data/atx_heading.html
@@ -1,9 +1,13 @@
-<h1>h1</h1>
-<h2>h2</h2>
-<h3>h3</h3>
-<h4>h4</h4>
-<h5>h5</h5>
-<h6>h6</h6>
+<h1 id="h1">h1</h1>
+<h2 id="h2">h2</h2>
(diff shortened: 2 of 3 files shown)


henriquemoody:
Looks like the problem is on Travis, not in my feature.

Seems PHP 5.2 is no longer available.

cebe:
> Seems PHP 5.2 is no longer available.

Yep, right: http://docs.travis-ci.com/user/build-environment-updates/2015-02-03/#PHP-VM

GrahamCampbell:
Yeh. They had to remove it because it was no-longer building on their system.

henriquemoody:
It's green now.

erusev:
Thanks. I'll have a look at this next weekend.

hkdobrev:
:+

naNuke:
Why not use ATX function instead of adding buch of code into text()?

henriquemoody:
@naNuke, I'm sorry but I don't know about ATX function. Can you elaborate?

erusev:
> This is more or less how GitHub Flavored Markdown works.

Is it described in [the specification](https://help.github.com/articles/github-flavored-markdown/)?

naNuke:
@henriquemoody line 468, sorry it was called atx in previous versions, now it's blockHeader()

henriquemoody:
@erusev, no it's not in the spec but If you take a look on all processed Markdown files rendered by GitHub you will he something like that right after `<hN>` tags:

```html
<a name="this-is-an-anchor" class="anchor" href="#this-is-an-anchor">
    <span class="octicon octicon-link"></span>
</a>
```

I don't think put these tags on processed Markdown files is a good idea because it's based on the GitHub's CSS, but with the ids we still have the reference and we can link the titles. It's useful when you want to write a table of contents or just create an link to the same document.

hkdobrev:
@erusev I don't think this is described in the spec. This is just how GitHub.com works. But it is really useful. Perhaps an extension? And then a community-driven wiki with a list of useful extensions?

henriquemoody:
I think better I close this.

GitHub Flavored Markdown only does it on pages and wiki, but it's not a standard (you can see that on comments and issue descriptions), besides, this parser may/must be used to render lots of files, if two different files have the same title (header) the `id` will not be generated correctly.

henriquemoody:
PS.: @erusev, I will keep my fork for a couple days, if you think this feature worth I will update it to reset the id counter every time something is rendered.

erusev:
@henriquemoody 

Thanks for elaborating on this. Since this is not part of the GFM spec, I'd have to agree with @hkdobrev and suggest that it would make more sense as an extension.

jmwright:
Has anyone created this extension, even if it hasn't been accepted into ParsedownExtra? I understand there are issues with it (same title and such), but I have a use case where I can control that if needed. If the extension doesn't exist I'll create it myself.

kminchev:
This being a feature for Extra repo was my first thought about this. I even recall this being discussed before.

jmwright:
Looks like it was discussed in issue #306 . Is there a reason that issue wasn't closed?

henriquemoody:
@jmwright, see https://github.com/erusev/parsedown/pull/285#issuecomment-73551128

sbrl:
Could this be added instead as an option that's disabled by default?

innaj:
I think a lot of people will need inline document navigation: What about optionally add an id-tag to the headers. Header "My Header " => creates a  "my-header" id inside the header tag.

decadence:
You can also create subclass without touching vendor files.
Laravel example.

```php
namespace App;

use Illuminate\Support\Str;

class HeadersParsedown extends \Parsedown
{
    protected function element(array $Element)
    {
        $headers = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'];

        $tag = $Element['name'];
        
        if (in_array($tag, $headers)) {
            $Element['attributes']['id'] = Str::slug($Element['text']);
        }

        return parent::element($Element);
    }
}
```

Issue #294, opened 2015-03-08
Release a new version compatible with PHP7


fonsecas72:
Hi, it seems to me that current version of parsedown is not compatible with PHP7 yet.
Can you please tag one more version? I think It's worth the effort. :)

erusev:
Sure,

@hkdobrev 

Since u did that for [Base](https://github.com/erusev/base), do u think u could do it for Parsedown as well?

hkdobrev:
I think Parsedown supports PHP 7 since version 1.5.3 and this could be closed now.

PR #295, opened 2015-03-08
Test on nightly releases of PHP (7.0) and HHVM

hkdobrev:
See https://github.com/erusev/parsedown/issues/294

See https://github.com/erusev/base/commit/12bf1c6d05e0ec89b1235078c6b5af3094422e0f#commitcomment-10090016

Nightly releases are allowed to fail since they are not stable and could fail in unpredictable ways.

New versions of PHP could be added as they are released.  I think they should not be in the allowed to fail part of the matrix then.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,16 @@
 language: php
 
 php:
+  - 7.0
   - 5.6
   - 5.5
   - 5.4
   - 5.3
   - hhvm
+  - hhvm-nightly
+
+matrix:
+  fast_finish: true
+  allow_failures:
+    - php: 7.0
+    - php: hhvm-nightly


.travis.yml
```diff
@@ -1,8 +1,15 @@
 language: php
 
 php:
+  - nightly
```
GrahamCampbell: Makes more sense to add 7.0 bellow 5.6...

hkdobrev: @GrahamCampbell That's the only upside...

fonsecas72:
JFTR, This does not closes #294 by it self, I just asked for a new tagged version.

hkdobrev:
@fonsecas72 Well, you're right, [there ar some changes](https://github.com/erusev/parsedown/compare/1.5.1...master) which are not in a tagged release. But you've asked a version compatible with PHP 7.

@erusev You could perhaps tag `1.5.2` even without merging this.

hkdobrev:
@fonsecas72 I've amended the commit and PR so it won't close the issue, even if merged.

fonsecas72:
@hkdobrev I already tried the master version with PHP7 and I realized that the tests were green. That's why I asked for the tagged version. But I think that both this PR and the new tag together makes more sense. Thank you :)

GrahamCampbell:
:-1: for this in it's current state. I see no point in testing on nighties. Seems largely overkill. Adding "7.0" because it's future proof is fine IMO.

.travis.yml
```diff
@@ -1,8 +1,15 @@
...
   - 5.6
   - 5.5
   - 5.4
   - 5.3
   - hhvm
+  - hhvm-nightly
+
+matrix:
```
fonsecas72: As I said to you in another place I think fast_finish: true should be added here.

cebe:
> :-1: for this in it's current state. I see no point in testing on nighties. Seems largely overkill. Adding "7.0" because it's future proof is fine IMO.

`nightly` is and alias for `7.0` for php.

fonsecas72:
According to this discussion ```nightly``` will be removed as soon as PHP7 is release
https://github.com/travis-ci/travis-ci/issues/2480
So I'd use ```7.0```

hkdobrev:
@fonsecas72 Thanks for the link. After reading the thread and few others, I agree that at the moment using `7.0` is more appropriate. But I don't agree with their decisions. I hope we would have PHP nightly builds just like `hhvm-nightly`, `ruby-head` and `jruby-head` even when PHP 7 is released.

Will update the PR.

/cc @GrahamCampbell

hkdobrev:
Any other issues with this?

/cc @erusev @GrahamCampbell @cebe @fonsecas72

fonsecas72:
fine by me :) :+1:

cebe:
looks perfect

GrahamCampbell:
:+1:

PR #302, opened 2015-03-28
PHP 7 compatibility fixes

GawainLynch:
Hi from the [Bolt CMS](http://bolt.cm) crew :wave: 

I'm working on getting us able to run on PHP 7 and found a compatibility issue with Parsedown:
```
PHP Fatal error: Function name must be a string
```

This PR resolves that by changing `$this->{$mystrin}` to use call_user_func() and call_user_func_array() instead.  

Passes the Parsedown PHPUnit tests on both PHP 5.x & 7

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -171,7 +171,7 @@ class Parsedown
 
             if (isset($CurrentBlock['incomplete']))
             {
-                $Block = $this->{'block'.$CurrentBlock['type'].'Continue'}($Line, $CurrentBlock);
+                $Block = call_user_func_array(array($this, 'block'.$CurrentBlock['type'].'Continue'), array($Line, $CurrentBlock));
 
                 if (isset($Block))
                 {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This reduces performance. :(

erusev:
Thanks. I'll have a more detailed look sometime next week. I wonder if the same can be achieved without adding function calls. Function calls, as @GrahamCampbell notes, have an effect on performance.

cebe:
isn't this a bug in php7? As far as I see there is no reason this should not work in 7.0.

GawainLynch:
@erusev they do, looping 1,000 PHPUnit runs on 5.6 there is a 2 millisecond average cost… However on PHP7 there is a 50 millisecond gain ;-)

GawainLynch:
@cebe your comment struck me with a "Hmm… why would Carsten object?" type of curiosity, so I set up a simple test case to prove *one* of us wrong (I can't be bothered with nerd-debates, evidence rules!)

Anyway, I can't reproduce it with a 10 line test case — with the same approach — *but*, in our tests it is failing! Dumping `$Element['handler']` (line 1405) I am seeing the valid string of "lines" when it's erroring under PHP 7 (built from cset 2fe6acd). So :+1: @cebe 

Long story short, I am going to pull this for now until I can better isolate what in Parsedown is triggering this.

erusev:
@cebe @GawainLynch Thanks.

GawainLynch:
Wow… I am going to hand in my programmers license! Our tests are running against v1.5.1 and cb8cc57742546648bb5461e66e35a92f2f89538b is not in a tagged release yet!

Nothing to see here, move along :grinning:

erusev:
I tagged the latest version in a 1.5.2 release.

PR #303, opened 2015-03-29
Make it possible to disable images

ctrlcctrlv:
Hello,

For privacy reasons, I do not want users adding images to markdown. For example, they could add 1px tracking GIF and it would be hard to spot. 

So I made a small change to make it possible to be disabled. If you use it make sure to also clean markup because they can still use ``.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -75,6 +75,16 @@ class Parsedown
 
     protected $urlsLinked = true;
 
+    function setImagesEnabled($imagesEnabled)
+    # Note: You should also set markupEscaped to disable images, otherwise people can still use <img> tags.
+    {
+        $this->imagesEnabled = $imagesEnabled;
+
+        return $this;
+    }
+
+    protected $imagesEnabled = true;
+
     #
     # Lines
     #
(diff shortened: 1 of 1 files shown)


Parsedown.php
```diff
@@ -75,6 +75,16 @@ function setUrlsLinked($urlsLinked)
 
     protected $urlsLinked = true;
 
+    function setImagesEnabled($imagesEnabled)
+    # Note: You should also set markupEscaped to disable images, otherwise people can still use <img> tags.
```
GrahamCampbell: probably a bad place to put a comment

hkdobrev:
@ctrlcctrlv Since this is something which affects only `inlineImage` and is not something popular it should probably be in an [extension](https://github.com/erusev/parsedown/wiki/Writing-Extensions). Other options like `breaksEnabled` and `urlsLinked` are used to enable/disable certain Markdown flavour-specific features from GFM for instance and `markupEscaped` is I think a popular request for use in Markdown-enabled user-generated data like comments.

---

P.S. Actually disabling inline images is suggested by @ersuev in the wiki:

> You can use the same approach for other elements types. This includes inline element types. For example, to disable Image elements you should override inlineImage.
> 
> &mdash; https://github.com/erusev/parsedown/wiki/Writing-Extensions#disable-element

erusev:
I'm closing this as I agree with @hkdobrev.

ctrlcctrlv:
OK, I understand, that does make sense. Thanks, I will use an Extension instead.

PR #309, opened 2015-04-15
Add support for omitting line breaks.

MeanEYE:
Omitting line breaks has no effect on browsers, and helps save precious few bytes where possible. This pull request adds `setLineBreaks` function which allows developers to specify whether they want line breaks included or not. By default line breaks _are_ included, which is current behavior of the parser.

Line breaks are removed only after closing tags and in few other place where they shouldn't create any problem. `pre`, `tt` and `code` tags and their content are unaffected.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -75,6 +75,15 @@ class Parsedown
 
     protected $urlsLinked = true;
 
+	function setLineBreaks($lineBreaks)
+	{
+		$this->lineBreaks = $lineBreaks;
+
+		return $this;
+	}
+
+	protected $lineBreaks = true;
+
     #
     # Lines
     #
@@ -275,11 +284,17 @@ class Parsedown
                 continue;
             }
 
-            $markup .= "\n";
+			if ($this->lineBreaks)
+			{
+				$markup .= "\n";
+			}
             $markup .= isset($Block['markup']) ? $Block['markup'] : $this->element($Block['element']);
         }
 
-        $markup .= "\n";
+		if ($this->lineBreaks)
+		{
+			$markup .= "\n";
+		}
 
         # ~
 
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -75,14 +75,14 @@ class Parsedown
 
     protected $urlsLinked = true;
 
-	function setLineBreaks($lineBreaks)
-	{
-		$this->lineBreaks = $lineBreaks;
+    function setLineBreaks($lineBreaks)
+    {
+        $this->lineBreaks = $lineBreaks;
 
-		return $this;
-	}
+        return $this;
+    }
 
-	protected $lineBreaks = true;
+    protected $lineBreaks = true;
 
     #
     # Lines
@@ -284,17 +284,17 @@ class Parsedown
                 continue;
             }
 
-			if ($this->lineBreaks)
-			{
-				$markup .= "\n";
-			}
+            if ($this->lineBreaks)
+            {
+                $markup .= "\n";
+            }
             $markup .= isset($Block['markup']) ? $Block['markup'] : $this->element($Block['element']);
         }
 
-		if ($this->lineBreaks)
-		{
-			$markup .= "\n";
-		}
+        if ($this->lineBreaks)
+        {
+            $markup .= "\n";
+        }
 
         # ~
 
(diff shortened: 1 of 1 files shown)


cebe:
I do not see any value in this. Saving "precious few bytes" is micro optimization which is unneeded. It also adds an ambigous method to the class as it can be easily confused with `setBreaksEnabled`.

MeanEYE:
Function name can easily be changed. Micro-optimization argument is however inaccurate. While for a single parsing one might save few to a hundred bytes, on pages which are viewed thousands of times this quickly amounts to much larger values. It might be negligible and a non-issue saying there's no benefit is not true. Any decent server hosting hundreds of sites with a fairly average view count will have significant increase in bandwidth use which could be used in a smarter way.

With availability of developer tools in all major browsers, reading page source is impractical. Adding line breaks for readability from page source not only wastes bandwidth but valuable CPU time as well. Not to mention use of double quotes in PHP is significantly slower due to parsing capabilities. 

Besides, am not arguing for the change of default behavior, rather having ability to optimize should one desire to do so. For my company, time-to-first-byte is important and considering page speed still affects ranking in Google I very much care for every millisecond I can save.

cebe:
> For my company, time-to-first-byte is important and considering page speed still affects ranking in Google I very much care for every millisecond I can save.

This actually argues against your implementation as you add an additional check that is run each time an element or block occurs in the markdown...

MeanEYE:
Which would be true if I didn't cache the output. Which I do. This parser adds code which slows down generation of output. In interest of not breaking expected behavior I added those checks and plan on relying on cache to solve problem I added. Ideally those new line characters should be removed as, in my opinion, they have no purpose unless code is using output in some editor.

smz:
IMHO this (HTML optimization/minification) is a concern better addressed at a next level of the page creation process, globally. Applying it for each HTML generation step (i.e. each component) is absolutely inneficient.

MeanEYE:
In my opinion @smz that makes little sense. Parsedown adds characters, which I would need to parse again and clean up in order to get better optimized page. At this point, I see that argument about having optimized code is waste of my time. I'll leave this pull request open for owner of this project to decide. If nothing else, I'll just have to maintain my own fork. All the best gentlemen.

smz:
Sorry, as I've said in the "edit" to my previous pots, it very much boils down to weather you'll be using CPU cycles for that or not. If yes, then you must consider if this represent a good "investment"  for subsequent steps.

MeanEYE:
Well @smz it comes with extra CPU cost because I added checks if option for making smaller code is enabled. Those checks are there purely for backwards compatibility. And that extra CPU time can be negated by caching output. Oh well. We'll see.

smz:
Thats, probably negigable: what matters are the CPU cycles used if/when your proposed option is active.

As an exemple, if you wish, look at one of my sites page: http://nuovaicona.org/

As you will see the genarted HTML is just on "line" of HTML code. This is a result you cannot achieve at the "component" level, but only through an optimizer for the "whole" HTML (and CSS, and Javascript) code.

If you use such an optimizer any previous partial optimization is worthless.

MeanEYE:
Well @smz I do that as well, my sites don't have any line breaks anywhere. But I don't wish to parse HTML code produced by _parsedown_ and remove extra stuff. That's just waste of resources. Instead it should be generated without newlines. I would be happy to modify my pull request which removes all the new line characters added to output code. That would be ideal solution as less operations would be needed and produced code would be smaller.

smz:
Sorry, I've just looked at your code and I agree: **adding**  '\n's to the generated markup is worthless and *it does consumes* CPU cycles. If simply removing them (as the default beheavior, without introducing any option) is accepatable, I will go for it: you have my vote. If instead a new option for the new behaviour is *politically needed*, then I'm unsure...

MeanEYE:
Thanks @smz. That was my initial thought, but it's not up to me to decide whether removing code is acceptable. That's something lead developer will have to decide on. I created PR with backwards compatibility, but it's easily transformed into version without option and without new line characters.

smz:
let's see...   :smile:

MeanEYE:
Taking too long to even reply on this. Am forking and maintaining my own version. Thanks.

PR #312, opened 2015-04-23
Fix check against $matches

garoevans:
Fixes inline reference links with int 0 as reference

The link [link][0] where [0] is set at the bottom of the md file current breaks and it's truthy value is false.


hkdobrev:
:+1: Nice catch!

[According to the docs](http://php.net/manual/en/language.types.boolean.php#language.types.boolean.casting) strings with "0" are evaluated to false. That's why we should always use `strlen` instead of casting to check for empty strings.

PR #316, opened 2015-05-07
set visibility of function text

DQNEO:
Thank you for a great tool.

I hope this fix helps it.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -21,7 +21,7 @@ class Parsedown
 
     # ~
 
-    function text($text)
+    public function text($text)
     {
         # make sure no definitions are set
         $this->DefinitionData = array();


DQNEO:
any problem?

DQNEO:
@erusev May I hear your opinion about this?

hkdobrev:
@DQNEO The accessor level of the methods in PHP is public by default. It depends on code style preference whether to explicitly state it or not.

DQNEO:
Thank you for the feedback.

It may be a preference, but I think PSR is a de-facto standard.

http://www.php-fig.org/psr/psr-2/#1-overview

> Visibility MUST be declared on all properties and methods; 

If you don't like this PR, please close it. I don't mind at all.

erusev:
i need to think about it a bit, but i'll probably convert all of the code to PSR eventually

hkdobrev:
> i'll probably convert all of the code to PSR eventually

:+1: :clap:

DQNEO:
:+1: 
I hope so.
I close this because one visibility change is not so useful.
Thank you for the response.

PR #328, opened 2015-06-12
Change PSR-0 to PSR-4

trupedia:
Since PSR-0 is deprecated, I changed it to PSR-4 and optimized the testsuite to use the composer generated autoload file (also added phpunit to the required-dev part).
But maybe it is even better just to use the classmap option in composer.json, because it's just a single class without a namespace at all.
Just my thoughts... 

Cheers


GrahamCampbell:
You should NOT move the file. That's a BC break. You can use PSR4 without moving any files.

.gitignore
```diff
@@ -0,0 +1,3 @@
+/vendor
+/.idea
+composer.lock
```
GrahamCampbell: cs

.gitignore
```diff
@@ -0,0 +1,3 @@
+/vendor
+/.idea
```
GrahamCampbell: NO. This should not go i here. This belongs in your global .gitignore file. People always do this wrong. :(

composer.json
```diff
@@ -12,7 +12,10 @@
             "homepage": "http://erusev.com"
         }
     ],
+    "require-dev": {
+      "phpunit/phpunit": "~4.0"
```
GrahamCampbell: Also wrong. This should NOT be added. Belongs in your global system install.

trupedia:
@GrahamCampbell thanks for the hints. Learnt a lot :)
Won't happen again!

.gitignore
hkdobrev: @GrahamCampbell :+1:

avindra:
+1 What's the possibility of switching to psr-4?

erusev:
@avindra it seems like a good idea, so i'd say that there's a good possibility

arcanedev-maroc:
Is there any progress about this issue ?

I've a suggestion about it and we can start applying the PSR-4 autoloading for the next major release `v2`:

 * The namespace will be `Parsedown\Parsedown` for the main class (Parsedown class will be inside the `src` folder).
 * Renaming the `test` folder to `tests` and of course we're going to use the PSR-4 `autoload-dev` and the namespace for the folder is going to be `Parsedown\Tests`.

Something like this:

```json
{
    "autoload": {
        "psr-4": {
            "Parsedown\\": "src/"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "Parsedown\\Tests\\": "tests/"
        }
    },
}
```

@erusev , can you create a `v2` branch to test it out ?

erusev:
@arcanedev-maroc i've been busy working on the Windows and Linux version of [Caret](http://caret.io). Hopefully, I'll have a look at this as soon as we release these versions later this month.

Issue #324, opened 2015-05-26
Documenting the functions in code


drdhaval2785:
It seems that the initial part of the class has comments which allow the user to see what happens.
But when the functions start, there is absolutely no comments.
Having it profusely commented would help the user to understand what is happenning in the code.

mastito03:
+1

PR #330, opened 2015-06-24
Added docblocks

Tjoosten:
referrer to #324


Parsedown.php
```diff
@@ -55,8 +78,12 @@ function setBreaksEnabled($breaksEnabled)
         return $this;
     }
 
-    protected $breaksEnabled;
-
+    /**
+     * setMarkupEscaped()
```
GrahamCampbell: :-1: for all these descriptions

GrahamCampbell:
:-1:

GrahamCampbell:
Most of these phpdocs are wrong or pointless.

Tjoosten:
Ok sorry u was trying to help

Parsedown.php
```diff
@@ -283,6 +324,13 @@ private function lines(array $lines)
     #
     # Code
 
+    /**
+     * blockCode()
+     *
+     * @param $Line
```
GrahamCampbell: this is no s unless you include the type

Parsedown.php
```diff
@@ -283,6 +324,13 @@ private function lines(array $lines)
...
+     * @param null $Block
```
GrahamCampbell: this type is wrong

Parsedown.php
```diff
@@ -309,6 +357,13 @@ protected function blockCode($Line, $Block = null)
         }
     }
 
+    /**
+     * blockCodeContinu()
+     *
+     * @param $Line
```
GrahamCampbell: and it's the same for all of it...

cebe:
there is no sense in adding docblocks that could be autogenerated by an IDE as they do not add any value to what is already visible from the code. If adding phpdoc comments they should contain useful description.

nblackburn:
Adding docblocks is a good idea but they at least need to be descriptive.

GrahamCampbell:
> Adding docblocks is a good idea but they at least need to be descriptive.

And actually have correct syntax. Nearly every param annotation here is missing the type so is incorrect syntax. In some cases the types are even wrong too.

PR #333, opened 2015-06-28
Added support of event listeners for blocks

BR0kEN-:
For each block, that converts to HTML, can be added an event listener - callback function with `$block` array as an argument. The one-time listeners are also supported.

Examples are available in modified `README.md`.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -261,19 +261,28 @@ class Parsedown
         # ~
 
         $markup = '';
+        $onetimeEvents = array();
 
         foreach ($Blocks as $Block)
         {
-            if (isset($Block['type']) and isset($this->EventListeners[$Block['type']]))
-            {
-                foreach ($this->EventListeners[$Block['type']] as $callback) {
-                  $callback($Block);
-                }
-            }
-
             if (isset($Block['hidden']))
             {
                 continue;
+            }
+
+            if (isset($Block['type']) and isset($this->EventListeners[$Block['type']]))
+            {
+                foreach ($this->EventListeners[$Block['type']] as $listener)
+                {
+                    $hash = spl_object_hash($listener['callback']);
+                    $executed = empty($onetimeEvents[$hash]);
+
+                    if ($executed)
+                    {
+                        $listener['callback']($Block);
+                        $onetimeEvents[$hash] = $listener['onetime'] and !$executed;
+                    }
+                }
             }
 
             $markup .= "\n";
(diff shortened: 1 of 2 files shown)


fesor:
This is pretty useful, but why just not add node visitor?

```
class MyTableProcessor extends NodeVisitor
{

    public function leaveTable(array $block) {
        $item = 0;

        // thead and tbody.
        foreach ($block['element']['text'] as &$section) {
             // tr.
             foreach ($section['text'] as &$rows) {
                  $rows['attributes']['class'] = $item++ % 2 ? 'odd' : 'even';
             }
        }
    }
}
```

or at least add visitTable/leaveTable events.

erusev:
@BR0kEN- 

It'd be helpful if you could briefly explain the problem that this solves, thanks.

BR0kEN-:
@fesor, what do you mean? As I know that library have no objects of such type. Please explain your idea.

@erusev, please [read the updated README.md](https://github.com/BR0kEN-/parsedown/blob/master/README.md#event-listeners) from my fork for understanding what event listeners does.

fesor:
@BR0kEN- it is true, without objects for nodes or event without intermediate structure node visitor is almost useless. But event listeners also. Anyway sometimes it will be useful to know that parser get some element and leave it. For example when parsing lists, i want to know when parser visit new list level and when it completes one.

BR0kEN-:
@fesor, I've made a thing that's useful for me and, maybe, another developers. And you're talking about another feature, not related to this. So, I'm not against of implementation, but not in this task.

erusev:
@BR0kEN- 

I'm more interested in specific scenarios that would benefit from the feature.

BR0kEN-:
@erusev, for example, you able to wrap any block with additional HTML or change it attributes or add some styles and javascript if some block present or, if necessary, modify content.

erusev:
@BR0kEN- 

Isn't this easy enough to do through extensions?

(see wiki for more info on extensions)

BR0kEN-:
@erusev, of course, but don't you say that creating additional objects for that is a good way?

fesor:
@BR0kEN- if you need to extend functionality for one single case, extensions is what should be used.

Events system will be very useful for some cases, but without visit/leave or start/complete events, which will allows you to track nested elements, is isn't cover most of possible use cases.

Sommerregen:
@erusev @aidantwoods Any chance to get this merged?

PR #334, opened 2015-07-14
Merge request for CommonMark compliance improvements

anthonywilliams:
Hi,

I've been using ParseDown for some Markdown processing, and wanted to use some CommonMark features that weren't supported.

I've forked the repository and made some changes that improve the conformance to CommonMark. It would be great if you could merge them into the main Parsedown repo.

In order to avoid any breaking changes to existing users, the new features are only enabled when explicitly requested. To enable the preservation of tabs use:

```
$pd=new Parsedown();
$pd->setExpandedTabs(false);
```

To enable the CommonMark HTML block processing use:

```
$pd->setCommonMarkHtmlBlocks(true);
```

I hope you find this a useful patch.

Anthony

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -412,7 +412,7 @@ class Parsedown
 
     protected function blockFencedCode($Line)
     {
-        if (preg_match('/^(['.$Line['text'][0].']{3,})[ ]*([\w-]+)?[ ]*$/', $Line['text'], $matches))
+        if (preg_match('/^(['.$Line['text'][0].']{3,})\s*([\w-]+)?\s*$/', $Line['text'], $matches))
         {
             $Element = array(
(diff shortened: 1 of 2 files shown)


aidantwoods:
@anthonywilliams any interest in updating this for the latest `master`? :)

There were some changes merged already that should mean we are now doing HTML blocks per CommonMark spec by default. 
From your changes, we probably also want to be preserving tabs by default (I can't foresee any major probalems with moving towards CommonMark in that regard). If it turns out that preserving tabs causes some unforeseen breakage then we can gate it behind strict mode instead of it being default, I think we should try matching spec first though :)

anthonywilliams:
I'll take a look in a couple of weeks. No free time just now.

PR #336, opened 2015-07-20
Use late static binding for Parsedown::instance()

hkdobrev:
Fixes erusev/parsedown-extra#67.

This introduces PHP 5.3+ late static binding to the Singleton pattern in Parsedown.
It will return an instance of Parsedown which inherits the class which called the `instance()` method rather than always returning instance of just `Parsedown`.

Tests are testing this feature with a test class which inherits from Parsedown.
Notice that calling `instance()` with the default arguments after an instance of `Parsedown` was already created, it will return it even though it is from just an instance of `Parsedown`. So this is fixing the problem just partially.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1475,7 +1475,7 @@ class Parsedown
             return self::$instances[$name];
         }
 
-        $instance = new self();
+        $instance = new static();
 
         self::$instances[$name] = $instance;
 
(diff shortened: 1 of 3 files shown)


mzeis:
@erusev Do you consider tagging a new version so `Parsedown` can be extended using an official release of `erusev/parsedown`? That would be great. Thanks!

erusev:
@mzeis done

mzeis:
Perfect, thanks!

PR #340, opened 2015-07-27
Improved extensibility by adding hooks which allow custom methods to be added without inheritance.

jaw-sh:
I have made Parsedown capable handling closures as a substitute for hard-defined methods.

There are two parts to this change.

1. A new protected property, `$elementClosures`.
2. A new method, `getElementMethod`, and 5 accessory methods that tie into this method. This will return one of three data types.
  - `void`, when the method is not found.
  - `Array` with the contents `[ 0 => $this, 1=> $methodName ]`.
  - `Closure`, where the requested method name exists as a key in `$elementClosures`.
3. Modifying the extremely large (and private!) methods `lines` and `line` to use the accessory methods when checking for a method / calling the method.

If you're wondering why this needs to be done, instead of just inheriting the Parsedown class and adding whatever the hell you feel like, here is an example.

I'm building a class called Eightdown, which inherits Parsedown, but itself is what's called a Facade. This package is meant for the Laravel framework, and the purpose of these Facades is to act as an API between specific libraries (Parsedown) and the application. The Facade provides a set of ambiguous methods which makes it very easy to have a single instance of code interact with multiple libraries. If I replaced Parsedown with something else, it would affect the actual application very little.

So to allow this to work with Parsedown, it needs to be able to handle supplied closures that extend or modify its feature set. This is what I have provided.

You can see an example of how the Facade interacts with the code here:
https://github.com/infinity-next/eightdown/blob/master/src/Traits/ParsedownExtensibility.php


rhukster:
I'm actually running into the exact same problem.  I'm the lead dev of the Grav CMS and we use Parsedown for our Markdown support, however we have lots of requests to provide our plugin support to also allow adding new functionality to Parsedown.  Extending is not an option at runtime, and I have a proof of concept working with **Inline** types without issue, however **Block** types are a problem because the **Line** method is 1) Private and 2) has `method_exists()` checks for `Continue` and `Complete` methods, which as you say are not going to find methods that have been added via closures and rely on a `__call()` to initiate them.

I propose a simpler less invasive solution:

1) Make Lines `protected` to let a subclass modify if needed (optional but why not)
2) Make 2 new `protected` methods: `isBlockContinueable($type)` and `isBlockCompleteable($type)`.  That way we can override these methods to look for a local state or check if the method_exists() as it does now.  Update `Lines` to use these.

I will submit a PR shortly...

hkdobrev:
Extensibility with protected methods was chosen over adding hooks. See #74. Certainly, @erusev opinion on that would be more useful.

PR #357, opened 2015-10-20
Parsedown library shouldn't be executable

donatj:
I was doing a little security inventory and found Parsedown.php to be executable. 

As it's not shell executable, it should not be.


erusev:
@hkdobrev could u share ur opinion on this

hkdobrev:
:+1: It should not be executable. It doesn't have a shebang and it's not structured to be executed directly. `php Parsedown.php` could make sense, but it would work with 0644 as well. https://github.com/erusev/parsedown/pull/275 purpose is to provide an executable CLI tool.

cebe:
:+1:

erusev:
thanks @donatj for the PR and thanks @hkdobrev and @cebe for the expressed opinions

PR #361, opened 2015-10-31
remove PHP 7.0 from allow failures in tests + improve readme

grogy:
From history in Travis-CI I see that this library is stable for use with PHP 7.0. And PHP 7 is coming soon.. :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,5 +12,4 @@ php:
 matrix:
   fast_finish: true
   allow_failures:
-    - php: 7.0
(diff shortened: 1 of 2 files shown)


hkdobrev:
:+1: 

Another thing is that `hhvm-nightly` is no longer supported for undisclosed time, so we could remove it from Travis tests.

grogy:
@hkdobrev thanks for your comment. I agree with your opinion about hhvm-nightly.

PR #372, opened 2015-12-17
Simple changes to make Parsedown more pluggable

rhukster:
An alternative and simpler approach to make Parsedown able to support plugins that extend functionality without requiring extending the class.  I outlined this approach in the comments of PR #340.

My approach is simpler and less invasive but allows for the ability to dynamically add methods to Parsedown by overriding the new `isBlockContinueable` and `isBlockCompleteable` methods.

In Parsedown itself, these will simply use the existing `method_exists()` check.  However in a 3rd party overridden class they can be extended to support other checks.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -121,7 +121,7 @@ class Parsedown
     # Blocks
     #
 
-    private function lines(array $lines)
+    protected function lines(array $lines)
     {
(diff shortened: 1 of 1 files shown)


rhukster:
Had an out of date branch.. will redo with current master...

PR #373, opened 2015-12-17
Simple changes to make Parsedown more pluggable

rhukster:
An alternative and simpler approach to make Parsedown able to support plugins that extend functionality without requiring extending the class.  I outlined this approach in the comments of PR #340.

My approach is simpler and less invasive but allows for the ability to dynamically add methods to Parsedown by overriding the new `isBlockContinueable` and `isBlockCompleteable` methods.

In Parsedown itself, these will simply use the existing `method_exists()` check.  However in a 3rd party overridden class they can be extended to support other checks.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -175,7 +175,7 @@ class Parsedown
                 }
                 else
                 {
-                    if ($this->isBlockCompleteable($CurrentBlock['type']))
+                    if ($this->isBlockCompletable($CurrentBlock['type']))
                     {
                         $CurrentBlock = $this->{'block'.$CurrentBlock['type'].'Complete'}($CurrentBlock);
                     }
@@ -216,7 +216,7 @@ class Parsedown
                         $Block['identified'] = true;
                     }
 
-                    if ($this->isBlockContinueable($blockType))
+                    if ($this->isBlockContinuable($blockType))
                     {
                         $Block['continuable'] = true;
                     }
@@ -245,7 +245,7 @@ class Parsedown
 
         # ~
 
-        if (isset($CurrentBlock['continuable']) and $this->isBlockCompleteable($CurrentBlock['type']))
+        if (isset($CurrentBlock['continuable']) and $this->isBlockCompletable($CurrentBlock['type']))
         {
             $CurrentBlock = $this->{'block'.$CurrentBlock['type'].'Complete'}($CurrentBlock);
         }
@@ -281,12 +281,12 @@ class Parsedown
     #
     # Allow for plugin extensibility
     #
-    protected function isBlockContinueable($Type)
+    protected function isBlockContinuable($Type)
     {
         return method_exists($this, 'block'.$Type.'Continue');
     }
 
-    protected function isBlockCompleteable($Type)
+    protected function isBlockCompletable($Type)
     {
         return method_exists($this, 'block'.$Type.'Complete');
     }
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -115,7 +115,7 @@ class Parsedown
     # Blocks
     #
 
-    private function lines(array $lines)
+    protected function lines(array $lines)
     {
         $CurrentBlock = null;
 
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1534,8 +1534,8 @@ class Parsedown
         'q', 'rt', 'ins', 'font',          'strong',
         's', 'tt', 'sub', 'mark',
         'u', 'xm', 'sup', 'nobr',
-        'var', 'ruby',
-        'wbr', 'span',
-        'time',
+                   'var', 'ruby',
+                   'wbr', 'span',
+                          'time',
     );
 }
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -175,7 +175,7 @@ class Parsedown
                 }
                 else
                 {
-                    if (method_exists($this, 'block'.$CurrentBlock['type'].'Complete'))
+                    if ($this->isBlockCompleteable($CurrentBlock['type']))
                     {
                         $CurrentBlock = $this->{'block'.$CurrentBlock['type'].'Complete'}($CurrentBlock);
                     }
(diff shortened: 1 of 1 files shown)


erusev:
@rhukster could u provide a simple use case example - perhaps, an simple extension that's built on top of the proposed architecture

rhukster:
Sure.. I am pasting some stripped down code to show the basics:

First i have my own class that extends Parsedown:

```php
<?php
namespace Grav\Common\Markdown;

class Parsedown extends \Parsedown
{
    use ParsedownGravTrait;

    public function __construct($page, $defaults)
    {
        $this->init($page, $defaults);
    }

}
```

Then I have a Trait that adds functionality for both Parsedown and ParsedownExtra. This provides a new `addBlockType` and `addInlineType` method that sets up the basics Parsedown needs to add the types.  Also you will see references to the new `isBlockContinueable` and `isBlockCompleteable` methods that override the default definition to take into account a local array.

```php
<?php
namespace Grav\Common\Markdown;

/**
 * A trait to add some custom processing to the identifyLink() method in Parsedown and ParsedownExtra
 */
trait ParsedownGravTrait
{
    use GravTrait;

    public $completeable_blocks = [];
    public $continueable_blocks = [];

    /**
     * Initialization function to setup key variables needed by the MarkdownGravLinkTrait
     *
     * @param $page
     * @param $defaults
     */
    protected function init($page, $defaults)
    {
        self::getGrav()->fireEvent('onMarkdownInitialized', new Event(['markdown' => $this]));
    }

    /**
     * Be able to define a new Block type or override an existing one
     *
     * @param $type
     * @param $tag
     * @param $continueable
     * @param $completeable
     */
    public function addBlockType($type, $tag, $continueable = false, $completeable = false)
    {
        $this->BlockTypes[$type] []= $tag;

        if ($continueable) {
            $this->continueable_blocks[] = $tag;
        }

        if ($completeable) {
            $this->completeable_blocks[] = $tag;
        }
    }

    /**
     * Be able to define a new Inline type or override an existing one
     *
     * @param $type
     * @param $tag
     */
    public function addInlineType($type, $tag)
    {
        $this->InlineTypes[$type] []= $tag;
        $this->inlineMarkerList .= $type;
    }

    protected function isBlockContinueable($type)
    {
        $continueable = in_array($type, $this->continueable_blocks) || method_exists($this, 'block'.$Type.'Continue');
        return $continueable;
    }

    protected function isBlockCompleteable($type)
    {
        $completeable = in_array($type, $this->completeable_blocks) || method_exists($this, 'block'.$Type.'Complete');
        return $completeable;
    }

    // For extending this class via plugins
    public function __call($method, $args)
    {
        if (isset($this->$method) === true) {
            $func = $this->$method;
            return call_user_func_array($func, $args);
        }
    }
}
```

Lastly is a Grav plugin that adds the Inline Color example from your FAQ and also a new **Notices** block type:

```php
<?php
namespace Grav\Plugin;

use \Grav\Common\Plugin;
use RocketTheme\Toolbox\Event\Event;

class MarkdownColorPlugin extends Plugin
{
    /**
     * @return array
     */
    public static function getSubscribedEvents()
    {
        return [
            'onMarkdownInitialized' => ['onMarkdownInitialized', 0]
        ];
    }

    public function onMarkdownInitialized(Event $event)
    {
        // Get markdown from the event
        $markdown = $event['markdown'];

        // Initialize Text example
        $markdown->addInlineType('{', 'ColoredText');

        // Add function to handle this
        $markdown->inlineColoredText = function($excerpt) {
            if (preg_match('/^{c:([#\w]\w+)}([^{]+){\/c}/', $excerpt['text'], $matches))
            {
                return array(
                    'extent' => strlen($matches[0]),
                    'element' => array(
                        'name' => 'span',
                        'text' => $matches[2],
                        'attributes' => array(
                            'style' => 'color: '.$matches[1],
                        ),
                    ),
                );
            }
        };

        // Initialize the Notices example
        $markdown->addBlockType('!', 'Notices', true, false);

        // Register the main block function
        $markdown->blockNotices = function($Line) {
            if (preg_match('/^(!{1,4}) (.*)/', $Line['text'], $matches))
            {
                $level = strlen($matches[1]);

                if ($level > 4)
                {
                    return;
                }

                $text = $matches[2];

                $Block = array(
                    'element' => array(
                        'name' => 'div',
                        'handler' => 'lines',
                        'attributes' => array(
                            'class' => 'alert'. $level,
                        ),
                        'text' => (array) $text,
                    ),
                );

                return $Block;
            }
        };

        // Register the continue function
        $markdown->blockNoticesContinue = function($Line, array $Block) {
            if ($Line['text'][0] === '!' and preg_match('/^![ ]?(.*)/', $Line['text'], $matches))
            {
                if (isset($Block['interrupted']))
                {
                    $Block['element']['text'] []= '';

                    unset($Block['interrupted']);
                }

                $Block['element']['text'] []= $matches[1];

                return $Block;
            }

            if ( ! isset($Block['interrupted']))
            {
                $Block['element']['text'] []= $Line['text'];

                return $Block;
            }
        };
    }
}
```

> NOTE: the Continue function doesn't work properly, i've not had time to debug it as i've been busy with other things all day today.  But it does not process the function. Before it never ever was run because `method_exists()` would not see it because it's made through a `__call()`

rhukster:
@erusev what do you think? It's such a small modification with no change in how Parsedown functions at all.  However it opens up Parsedown to be much more extensible.

erusev:
@rhukster it looks good, in fact, it's quite impressive that you've managed to achieve this kind of extensibility using with so few changes to the code. Can we just change `Completeable` to `Completable`. It'd also be great if you could create a simple wiki page that explains this alternative approach to extensibility.

rhukster:
Fixed the spelling errors.. Seems I spelt both "continuable" and "completable" wrong :(  Anyway fixed now.

Also, yes I could add something to the wiki page showing how to extend Parsedown in a plugin-friendly way. Probably early next week though.  Thanks!

PR #377, opened 2016-01-06
Update composer.json requirements

### LINKED ISSUE #374: Parse error
Parse error: syntax error, unexpected T_STATIC, expecting T_STRING or T_VARIABLE or '$' in my/path on line 1492

Link 1492: $instance = new static();

PhrozenByte:
Fixes #374

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Better Markdown Parser in PHP
 
 ### Installation
 
-Include `Parsedown.php` or install [the composer package](https://packagist.org/packages/erusev/parsedown).
+Include `Parsedown.php` or install [the composer package](https://packagist.org/packages/erusev/parsedown). Parsedown requires PHP 5.3+
 
 ### Example
 
--- a/composer.json
+++ b/composer.json
@@ -12,6 +12,9 @@
             "homepage": "http://erusev.com"
         }
     ],
+    "require": {
+        "php": ">=5.3.0"
+    },
     "autoload": {
         "psr-0": {"Parsedown": ""}
     }


hkdobrev:
:+1:

erusev:
why does this fix #374

PhrozenByte:
parsedown requries PHP 5.3+ (as stated in the `README.md`), this simply prevents (composer) installations with older PHP versions.

hkdobrev:
BTW `composer` won't run on < PHP 5.3.3 anyway, but this change shows the minimum version on packagist as well so you know if you can use it in your environment.

cebe:
composer needs php 5.3.3 or higher so there is not really a point in adding it.

cebe:
@hkdobrev ha, same thought in the same second :)

hkdobrev:
@cebe If it's missing, one could not assume if it would work everywhere where Composer works or if it was just omitted. Explicit it better than implicit.

cebe:
fine with me. But we had the discussion already here: https://github.com/erusev/parsedown/pull/38

hkdobrev:
I thought of #38 as well, but it's a bit different with PHP 5.2. However in both cases, I'd always prefer to state the requirement than leave it to the user to test and figure it out.

PhrozenByte:
`composer.json` isn't necessarily limited to the `composer` cli, it can also be interpreted by other tools and services as well, like packagist. If there is a explicit requirement, specify it explicitly. This is common sense in like every package/dependency manager (e.g. `dpkg` and `rpm`). Never rely on that one of your dependencies (`composer` in this case) provides one of your direct dependencies (PHP 5.3+) implicitly, otherwise the dependency is completely out of your control.

erusev:
> composer.json isn't necessarily limited to the composer cli, it can also be interpreted by other tools and services as well

that's a good point

PhrozenByte:
Thank you @erusev :+1:

PR #380, opened 2016-01-20
Support for opening links in new tabs/windows

JeanValjean2:
I needed a simple Markdown syntax modification for potentially opening links in new tabs/windows.
If you add a '+' sign just before the http link there's a "target='_blank'" attribute added to a HTML A element.
Yes, it's a Markdown syntax modification...


erusev:
Thanks for this, but since it's not part of the specification, its place is in an extension.

Issue #378, opened 2016-01-18
Bug with multibyte text inside li tag


dotzero:
``` php
$text = "
* one.
* two
* раз.
* два
";

$md = new Parsedown();
var_dump($md->text($text));
```

Parsed into..

``` html
<ul>
<li>one.</li>
<li>two</li>
<li>ра/p></li>
<li>д�/p></li>
</ul>
```

erusev:
are u able to reproduce the issue at http://parsedown.org/demo

dotzero:
No,
but i think it cause on our server using 
`mbstring.func_overload = 7` and i unable to change it.

erusev:
can someone review this? i'm a bit overwhelmed with work these days

cebe:
if `mbstring.func_overload` is set on the server, this may cause a lot of problems in other places of Parsedown, not only this issue in particular. Parsedown relies on substr() and strlen() a lot to work on byte level. By overriding the functions with their mb_ counter parts this may break a lot of code in parsedown as these functions now work on characters instead of bytes.
To properly fix this, a wrapper function could be created that always behaves the correct way, regardless of the ini setting:

```php
function my_strlen($string) {
    if (function_exists('mb_strlen')) {
        return mb_strlen($string, '8bit');
    } else {
        return strlen($string);
    }
}
```

Similar functions need to be created for substr and maybe others too.

PhrozenByte:
parsedown shouldn't endorse people to use highly discouraged "features" (I've no idea why PHP devs decided to include this mess...) like mbstring function overloading by explicitly supporting it. Either leave it the way it is and tell people to fix their PHP configuration or let parsedown explicitly die when a malicious configuration is detected:

```php
if (@ini_get('mbstring.func_overload') >= 2) {
    throw new ErrorException('You must disable mbstring function overloading');
}
```

aidantwoods:
Since it didn't appear to be possible to reproduce this issue, I'm going to close this one – please reopen if you can reproduce this with the latest version though. If the problem is only due to PHP configuration I'm not really sure there's much that can be done by Parsedown.

PR #381, opened 2016-01-25
Hotfix multibyte lists #378

dotzero:
Hotfix multibyte lists #378

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1459,9 +1459,8 @@ class Parsedown
             $markup = $trimmedMarkup;
             $markup = substr($markup, 3);
 
-            $position = strpos($markup, "</p>");
-
-            $markup = substr_replace($markup, '', $position, 4);
+            $count = 1;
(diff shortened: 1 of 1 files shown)


Parsedown.php
```diff
@@ -1459,9 +1459,8 @@ protected function li($lines)
             $markup = $trimmedMarkup;
             $markup = substr($markup, 3);
 
-            $position = strpos($markup, "</p>");
-
-            $markup = substr_replace($markup, '', $position, 4);
+            $count = 1;
+            $markup = str_replace('</p>', '', $markup, $count);
```
erusev: is the `$count` variable necessary

dotzero: yes, cause arg can pass a variable by reference only.

aidantwoods:
Hi, thanks for the PR!

There was some similar work done in #513, so I'm going to close this in favour of that since it aims to bring multibyte support to all the structures in Parsedown (I think it's probably about time we patch this everywhere). I hope this makes sense :)

PR #388, opened 2016-02-17
Allow user to set a relative path

bct9321:
I needed a way to access images in different folders with the latest parsedown version. You can now set a relative path to add to image tags and not have to use absolute paths.


duckbrain (approved):
I've tested these changes they provide just what I need. Thanks @bct9321!

NinjaPewPew:
It worked, well done. Too bad this was never merged.

PR #403, opened 2016-06-06
Avoid XSS via links and images

JuliusPC:
Maybe there’s a better way doing this but it works...

Some test cases:

```
[Test](javascript:alert(\'HACK3D!!!\')) [xss](http://"onmouseover="alert(1)

[xss-mailto](mailto:test@example.org"onmouseover="alert(1))

![Test](image.jpeg"onmouseover="alert(1))
```

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -42,7 +42,7 @@ class Parsedown
         $markup = trim($markup, "\n");
 
         # avoid XSS in links via javascript:dosomething()
-        $markup = str_replace('<a href="javascript:', '<a href="', $markup);
+        $markup = str_ireplace('<a href="javascript:', '<a href="', $markup);
 
         return $markup;
     }


PhrozenByte:
It's Markdown, the whole point of Markdown is to use HTML when the Markdown syntax isn't sufficient, so what's the point of using hacks like this if you can simply add a `<script>` element?

JuliusPC:
Not in Parsedown: If you include `$Parsedown->setMarkupEscaped(true);` HTML gets escaped:  
`<script>alert('hacked!');</script>`→ `&lt;script&gt;alert('hacked!');&lt;/script&gt;`

PhrozenByte:
Ah, I see, my fault. In this case I've said nothing :+1:

Parsedown.php
```diff
@@ -1209,7 +1212,7 @@ protected function inlineLink($Excerpt)
 
         if (preg_match('/^[(]((?:[^ ()]|[(][^ )]+[)])+)(?:[ ]+("[^"]*"|\'[^\']*\'))?[)]/', $remainder, $matches))
         {
-            $Element['attributes']['href'] = $matches[1];
+            $Element['attributes']['href'] = htmlspecialchars($matches[1], ENT_QUOTES, 'UTF-8');
```
aidantwoods: This change conflicts with line 1249 at present (where some less rigorous escaping occurs).

I'd suggest moving line 1249 into the `else` block immediately above it (your changes cause double HTML escaping in some cases). That should solve the test failures you're getting at present.

Alternatively, I guess you could swap the less rigorous escaping on line 1249 with this method (though I'm not sure if that will cause any additional conflicts)

tillkruss:
Is this still happening?

PR #404, opened 2016-06-17
Remove & from being replaced with &amp; in inline links

33mhz:
Currently the `&` is converted into `&amp;` in an inline link

`[this is my site](https://example.com/site?param1=1&param2=2)`

is converted into

`<a href="https://example.com/site?param1=1&amp;param2=2">this is my site</a>`.

I think it should be fine to simply remove that.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1243,7 +1243,7 @@ class Parsedown
             $Element['attributes']['title'] = $Definition['title'];
         }
 
-        $Element['attributes']['href'] = str_replace(array('&', '<'), array('&amp;', '&lt;'), $Element['attributes']['href']);
+        $Element['attributes']['href'] = str_replace('<', '&lt;', $Element['attributes']['href']);
 
         return array(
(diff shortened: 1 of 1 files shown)


PhrozenByte:
The current behavior is completely fine, you *must* escape `&` in HTML attributes.

33mhz:
@PhrozenByte Haha. Of course. I see, in case I need the link to have a parameter that happens to collide with an HTML entity.

Unfortunately, the problem I was running into is that Parsedown is double-encoding `&` in hrefs. So my fix still stands. The `&` is already escaped elsewhere.

PhrozenByte:
I cannot reproduce the double-encoding with http://parsedown.org/demo

33mhz:
@PhrozenByte neither can I. Strangely, the same contents, same code, same versions work on my HTTP IP testing site, and don't work on my HTTPS domain site. I'll have to investigate what that is about, but it isn't the code. Thanks.

PR #406, opened 2016-06-24
Create .gitattributes

adrilo:
When using this library, all the files related to tests can be ignored. Tests are only useful when working on the library itself.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# Ignore all tests for archive
+/test               export-ignore
+/.gitattributes     export-ignore
+/.travis.yml        export-ignore
+/phpunit.xml.dist   export-ignore


hkdobrev:
👍 This would speed up download times when installing it through Composer.

stof:
👍 for this.

The class alias created for the `PHPUnit\Framewor\TestCase` confuses PHPStorm as it detects a double class declaration (the class alias in yoru testsuite and the actual declaration in PHPUnit), and then refuses to provide autocompletion for methods in it.
Removing the testsuite from the archive generated by github would solve this.

erusev:
Does anyone know why the build errors?

stof:
@erusev the installation of HHVM failed. This should be reported to Travis, as it is an issue on their side.

aidantwoods:
https://github.com/travis-ci/travis-ci/issues/7712

erusev:
Thanks!

aidantwoods:
I fixed it in #499 :)

stof:
@erusev could you make a new release ?

erusev:
@stof done

PR #409, opened 2016-07-16
Don't create pair tags without content

integer:
This commit fix empty `<p>` tags after parsing markdown with custom extensions like this: https://gist.github.com/integer/05dea90948c4e96a254bb2508151d1ef


aidantwoods:
Empty tags might sometimes be intended, instead when #576 is finished you could traverse the AST and specifically filter out empty paragraphs in your extension though :)

PR #426, opened 2016-09-06
Prefix relative images with server path

zorker:


PhrozenByte:
No! This (a) breaks existing setups (i.e. breaks BC) and (b) breaks future setups. You can't predict in a library how applications route requests. This is something the application has to deal with.

tovic:
Checking relative URL is not that easy.

https://github.com/mecha-cms/mecha-cms/blob/176ae133821cfd3be4cb8b6c96d0be197fd3778a/engine/kernel/converter.php#L29&L39

Issue #407, opened 2016-07-13
According to the web component spec, a dash is required, which breaks md parsing


rodsimpson:
If you are using web components in your html / md, you must have a dash.  For example:

Invalid: 
`<mycomponent></mycomponent>`

Valid: 
`<my-component></my-component>`

The former case stops further parsing of markdown. For example:

Works:
`<mycomponent></mycomponent>`
`# header`

Doesn't work:
`<my-component></my-component>`
`# header`

In the second case, # header is not converted to an H1 tag, but it should be.

aarakast:
The following seems to fix it:

Line 676:
`-         if (preg_match('/^<(\w*)(?:[ ]*'.$this->regexHtmlAttribute.')*[ ]*(\/)?>/', $Line['text'], $matches))`
`+         if (preg_match('/^<(\w[\w-]*\w)(?:[ ]*'.$this->regexHtmlAttribute.')*[ ]*(\/)?>/', $Line['text'], $matches))`

Dito on line 1277:
`-        if ($Excerpt['text'][1] !== ' ' and preg_match('/^<\w*(?:[ ]*'.$this->regexHtmlAttribute.')*[ ]*\/?>/s', $Excerpt['text'], $matches))`
`+        if ($Excerpt['text'][1] !== ' ' and preg_match('/^<\w[\w-]*\w(?:[ ]*'.$this->regexHtmlAttribute.')*[ ]*\/?>/s', $Excerpt['text'], $matches))`

This changed regex allows [custom element names](https://html.spec.whatwg.org/multipage/scripting.html#valid-custom-element-name) such as `<valid-tag>` but not `<-notvalid>` or `<notvalid->`. I have not tested the code against all the other unicode characters mentioned in above link.

PhrozenByte:
@pablotheissen: Don't make it overcomplicated, Parsedown is no HTML validator, it's just about matching HTML elements, not about validating them. So, just add the `.` next to the `-` as allowed char and this should be good to go.

tovic:
What about namespaces? Example `<foo:bar baz:qux="xyz">`

~~~ .javascript
/<[\w:.-]+(\s[^<>]*?)?>/
~~~

aidantwoods:
Provided that you include a newline between the HTML and the markdown following it this should work as expected in the next release :)

PR #429, opened 2016-09-22
Support html tags containing dashes

pablotheissen:
Made parsedown compatible with html-tags containing dashes.
see https://github.com/erusev/parsedown/issues/407#issuecomment-248833563


micbenner:
Can this get merged please

PR #430, opened 2016-09-23
Fix em and strong regex bug – failure to correctly recognise double nested escape sequences

aidantwoods:
Em and Strong regex were unable to recognise double nested escape sequences.
The previous strong regex didn't find the following to be a match
https://regex101.com/r/rH9oJ6/2
The updated regex recognises it
https://regex101.com/r/bC4xE5/4

Similarly, the em regex did not recognise this as a match
https://regex101.com/r/jS1lH9/2
But the updated regex again correctly matches
https://regex101.com/r/aU7oO1/2


rdelepine:
Not match this : `**italic *bold\*it* italic** normal **italic**`

aidantwoods:
Was that match found with the code, or the link I gave?

Apologies on the link – I naively copy pasted the regex used directly, without removing the backslashes that PHP itself will treat as literal ;p

Link should have been https://regex101.com/r/bC4xE5/3

PhrozenByte:
Unfortunately the regex considers just a single escape char, thus a escaped escape char isn't treaded correctly (i.e. you don't have to search for a single escape char, but any uneven number of escape chars). Thus the following is matched wrong:
https://regex101.com/r/3aZlVj/1

aidantwoods:
@PhrozenByte I think you might be touching on a series of escaping problems within parsedown in various places ;p (#434) 
I can address escapable escapes in the regexes above (e.g. this should fix the one you sent https://regex101.com/r/Y3rFDL/1), though this will probably be the first of many changes that need to be made to escaping

e.g.
screen shot 2016-10-04 at 21 46 39

PhrozenByte:
You have to start somewhere :laughing: However, @erusev should decide whether this "escape escaping chars" issue should be addressed now or not.

Issue #100, opened 2014-02-13
Specify start attribute on ordered lists


hkdobrev:
According to the Markdown spec any number with a following dot could start an ordered list.

Like so:

``` md
5. fifth
6. sixth
7. seventh
```

[In the HTML5 spec there is a `start` attribute on `ol` elements](http://www.w3.org/TR/html5/grouping-content.html#attr-ol-start) which could indicate the start of the list.

So this should produce something like:

``` html
<ol start="5">
<li>fifth</li>
<li>sixth</li>
<li>seventh</li>
</ol>
```

---

Other implementations which do that:
- [cheapskate](https://github.com/jgm/cheapskate)
- [pandoc](https://github.com/jgm/pandoc)

---

Thanks should go to @wkpark for revealing it in #99.

rmjarvis:
+1 on this feature request.  I am constantly bitten by this "feature" of markdown.  e.g. if someone makes a list of 5 things in an issue, and I want to respond to the 4th one, I'll try to copy it out for a response:
```
> 4. The fourth item that someone talked about

This is already implemented.  See this issue...
```
Shows up as:

> 4. The fourth item that someone talked about

This is already implemented.  See this issue...

It would be nice to have *some* way to force the numbering to start where I want it, not at 1.

erusev:
@rmjarvis that's a good point, i'll se what i can do

aidantwoods:
@erusev just taking a quick glance at the source for this issue

For the following section of code (which creates tags) 
```
        if (isset($Element['text']))
        {
            $markup .= '>';
            if (isset($Element['handler']))
            {
                $markup .= $this->{$Element['handler']}($Element['text']);
            }
            else
            {
                $markup .= $Element['text'];
            }
            $markup .= '</'.$Element['name'].'>';
        }
```
Would the following change cause any issues?
```
$markup .= '</'.$Element['name'].'>';
``` 
to 
```
$markup .= '</'.preg_replace('/[ ][^ ]+/', '', $Element['name']).'>';
```
This should just strip everything after a space (if any, and including the space) out of the tag name (which a tag shouldn't have in it anyway at the close).

That change would allow you to change the blockList function from
```
protected function blockList($Line)
    {
        list($name, $pattern) = $Line['text'][0] <= '-' ? array('ul', '[*+-]') : array('ol', '[0-9]+[.]');
        if (preg_match('/^('.$pattern.'[ ]+)(.*)/', $Line['text'], $matches))
        {
[...]

PR #431, opened 2016-09-26
Allow parsedown to specify list start attribute

aidantwoods:
(I think this should work)
Allow parsedown to specify list start attribute (see: https://github.com/erusev/parsedown/issues/100#issuecomment-249729602)

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -503,6 +503,7 @@ class Parsedown
     protected function blockList($Line)
     {
         list($name, $pattern) = $Line['text'][0] <= '-' ? array('ul', '[*+-]') : array('ol', '[0-9]+[.]');
+        
         if (preg_match('/^('.$pattern.'[ ]+)(.*)/', $Line['text'], $matches))
         {
             $Block = array(
@@ -513,14 +514,16 @@ class Parsedown
                     'handler' => 'elements',
                 ),
             );
+            
             if($name === 'ol') 
             {
-                $list_num = stristr($matches[0], ".", true);
-                if($list_num !== '1')
+                $listStart = stristr($matches[0], ".", true);
+                if($listStart !== '1')
                 {
-                    $Block['element']['attributes'] = array('start' => $list_num);
+                    $Block['element']['attributes'] = array('start' => $listStart);
                 }
             }
+            
             $Block['li'] = array(
                 'name' => 'li',
                 'handler' => 'li',
@@ -528,7 +531,9 @@ class Parsedown
                     $matches[2],
                 ),
             );
+            
             $Block['element']['text'] []= & $Block['li'];
+            
             return $Block;
         }
     }
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -517,7 +517,8 @@ class Parsedown
 
             if($name === 'ol') 
             {
-                $listStart = stristr($matches[0], ".", true);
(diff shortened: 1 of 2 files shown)


aidantwoods:
Okay this looks good.

There is 1 failure (which is entirely by design – to add the new feature)

screen shot 2016-09-27 at 01 25 15

PhrozenByte:
@erusev: Increases CommonMark compliance, see http://spec.commonmark.org/0.26/#start-number (however, I didn't check the suggested code changes)

aidantwoods:
I've removed a change I made that added a feature that already exists ;p
Have rewritten to use the existing functionality – guess that's what I get for skimming the code ;p

Checks still look good – still receiving (the expected) single failure

PhrozenByte:
Can you please update the test? You just have to edit `test/data/ordered_list.html` :smiley:

aidantwoods:
Okay, this should be good (including tests).
Had to modify the test via file upload, as GitHub auto-appends a new line when saving o.0

erusev:
Wouldn't it work if instead of line 505, you add the brackets on 506 and keep 510 unchanged?

image

aidantwoods:
@erusev Sort of – the dot still needs to be excluded somehow, whether that's by grabbing all the numbers in the capture group, or by chopping off the last character (the dot).

These would be the changes for the former method
screen shot 2016-10-04 at 21 20 35

erusev:
You are right. 

How about this - 505 and 510 stay the same, but in the if at 516, work with the existing match and split it by "." to get the number.

image

aidantwoods:
@erusev in-fact ignore my previous comment – that wouldn't work, since the number needs to be known before inserting it (to check if it is 1)

It would actually need to be more like this (with the string splitting to get the dot)

screen shot 2016-10-04 at 22 00 49

erusev:
This looks much better.

Could you update the pull request?

Thanks.

aidantwoods:
@erusev Done – as written in the screenshot, bar the `explode` function – build tests indicate PHP 5.3 had trouble navigating the array as it was returned by `explode` (guessing I'd have to use an intermediary variable – which impacts readability slighly IMO). So I've just used a `preg_replace` in its place to grab all the digits before the first `.`.

erusev:
`preg_replace` is very slow, though - check this out: http://stackoverflow.com/questions/1744956/explode-and-get-a-value-in-one-line-of-code

aidantwoods:
@erusev looks like `stristr` is working as expected 😃 any other changes?

erusev:
Great, a few more :) 

1. We don't use `_` as a word separator
2. No need to remove empty lines
3. It'd be more consistent if the if block is surrounded by empty lines

aidantwoods:
@erusev Syntax has been adjusted ;-)

erusev:
Could you handle 2 more :)

1. Single quotes on `"."`
2. Empty line before the nested `if`

aidantwoods:
@erusev Okay, that should be done now? 😜

erusev:
Nice, that's perfect 🎉

aidantwoods:
@erusev Nice! 😃 Guess it's probably safe to close https://github.com/erusev/parsedown/issues/100 and its duplicates now

erusev:
Yep :)

Issue #358, opened 2015-10-24
Parse inlineLink incorrectly if the link text is an URL


hoducha:
Test data:

```
[http://example.com](http://example.com)
```

Expected result:

``` html
<a href="http://example.com">http://example.com</a>
```

Actual result:

``` html
<a href="http://example.com"><a href="http://example.com">http://example.com</a></a>
```

petermolnar:
Duplicate of https://github.com/erusev/parsedown-extra/issues/77 ?

petermolnar:
This could be the reason behind https://github.com/erusev/parsedown-extra/issues/77

francisbesset:
@petermolnar This issue can be reproduced with parsedown only.

aidantwoods:
I've added made a pull request https://github.com/erusev/parsedown/pull/433 to address this issue

aidantwoods:
Fixed in #433

Issue #432, opened 2016-10-01
Link parse issue


jinliming2:
I wrote this in my markdown file:

``` markdown
[https://www.example.com](https://www.example.com/?page=5&load=4)
```

And then, I got this by parsedown:

``` html
<p><a href="https://www.example.com/?page=5&amp;load=4"><a href="https://www.example.com">https://www.example.com</a></a></p>
```

But what I wanted is:

``` html
<p><a href="https://www.example.com/?page=5&load=4">https://www.example.com</a></p>
```

kminchev:
Parses fine in http://parsedown.org/demo

PhrozenByte:
@kminchev: It does not.

@772807886: Duplicate of #358.

Note: You can disable auto-linking by setting `$Parsedown->setUrlsLinked(false)`.

PR #433, opened 2016-10-01
Fix Issue #358 – preventing double nested links

aidantwoods:
1. Add the ability for a parsed element to enforce that the line handler not parse any (immediate) child elements of a specified type.
2. Use 1. to allow parsed Link elements to tell the line handler not to parse any child Links or Urls where they are immediate children.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -987,7 +987,7 @@ class Parsedown
     # ~
     #
 
-    public function line($text, $non_nestables=array())
+    public function line($text, $nonNestables=array())
     {
         $markup = '';
 
@@ -1005,7 +1005,7 @@ class Parsedown
             {
                 # check to see if the current inline type is nestable in the current context
 
-                if ( ! empty($non_nestables) and in_array($inlineType, $non_nestables))
+                if ( ! empty($nonNestables) and in_array($inlineType, $nonNestables))
                 {
                     continue;
                 }
@@ -1033,9 +1033,9 @@ class Parsedown
 
                 # cause the new element to 'inherit' our non nestables
 
-                foreach ($non_nestables as $non_nestable)
+                foreach ($nonNestables as $non_nestable)
                 {
-                    $Inline['element']['non_nestables'][] = $non_nestable;
+                    $Inline['element']['nonNestables'][] = $non_nestable;
                 }
 
                 # the text that comes before the inline
@@ -1197,7 +1197,7 @@ class Parsedown
         $Element = array(
             'name' => 'a',
             'handler' => 'line',
-            'non_nestables' => array('Url', 'Link'),
+            'nonNestables' => array('Url', 'Link'),
             'text' => null,
             'attributes' => array(
                 'href' => null,
@@ -1425,14 +1425,14 @@ class Parsedown
         {
             $markup .= '>';
 
-            if (!isset($Element['non_nestables'])) 
+            if (!isset($Element['nonNestables'])) 
             {
-                $Element['non_nestables'] = array();
+                $Element['nonNestables'] = array();
             }
 
             if (isset($Element['handler']))
             {
-                $markup .= $this->{$Element['handler']}($Element['text'], $Element['non_nestables']);
+                $markup .= $this->{$Element['handler']}($Element['text'], $Element['nonNestables']);
             }
             else
             {
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -987,7 +987,7 @@ class Parsedown
     # ~
     #
 
-    public function line($text)
+    public function line($text, $non_nestables=array())
     {
         $markup = '';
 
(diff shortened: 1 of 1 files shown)


aidantwoods:
If we like 1. as a feature, then I can probably extend the functionality such that (disallowed) nesting can be asserted past a depth of one.

i.e. So although this PR fixes the original issue, cases such as:
```
[http://example.com](http://example.com)
```
It won't fix cases such as:
```
[*http://example.com*](http://example.com)
```
without being able to assert further than an immediate depth.

This is because the link can only currently assert nested exclusions for direct children. In the second example, the direct child of the Link is an italic element, and the inline Url is a child of the italic element (child of depth 2 relative to the Link).

aidantwoods:
I've build a little on the functionality as hinted in the above comment.

Inside the element array, one can now specify: 
`'non_nestables' => array('Url', 'Link'),` 
or 
`'non_nestables' => array(['Url'], ['Link']),` 
or mixed: 
`'non_nestables' => array('Url', ['Link']),` 
or the identical: 
`'non_nestables' => array('Url', array('Link')),` 

If you wish to only assert non nesting depth a finite number of times simply use the array syntax, and specify the depth assertion as the second array value.

e.g. the following will assert Link or Url may not be nested within a Link, at any depth, but Code may be inserted if greater than depth two:
screen shot 2016-10-02 at 17 32 33

So for the following data:
screen shot 2016-10-02 at 17 32 43

We achieve the following output:
screen shot 2016-10-02 at 17 32 57

Where inline Url is not ever allowed, but code is allowed past a depth of two

aidantwoods:
@PhrozenByte @erusev any thoughts on the latest changes in this PR?

Just noticed another duplicate issue (#432) report popped up for this right before my PR, so looks like this fix is a much needed update

PhrozenByte:
All in all: Looks great! :+1:

However, I don't see a use case for specifying a maximum depth for non nestable types. I suggest removing the "array syntax" (i.e. `'non_nestables' => array('Url', 'Link')` is the only supported syntax) and assuming that non nestable types are in effect for an infinite depth.

aidantwoods:
@PhrozenByte That makes sense to me!
Yeah, I couldn't immediately think of any useful cases for max nesting either – was an easy add on with the inheritance change though.

I've removed it for now, but if it's ever useful then it's easy to re-add if it's ever needed (and forward compatible with the current string syntax etc..)

It does simplify the code considerably not having the max depth and mixed syntax ;-)

Parsedown.php
```diff
@@ -1003,6 +1003,13 @@ public function line($text)
 
             foreach ($this->InlineTypes[$marker] as $inlineType)
             {
+                # check to see if the current inline type is nestable in the current context
+
+                if (in_array($inlineType, $non_nestables))
```
PhrozenByte: How about `if (!empty($non_nestables) && in_array($inlineType, $non_nestables))`? `in_array()` really is *very* expensive, even with empty arrays.

Parsedown.php
```diff
@@ -1040,26 +1031,11 @@ public function line($text, $non_nestables=array())
...
-                    elseif (is_array($non_nestable) and ! isset($non_nestable[1]))
-                    {
-                         $Inline['element']['non_nestables'][] = array($non_nestable[0]);
-                    }
-                    # string syntax, so depth is indefinite
-                    elseif ( ! is_array($non_nestable))
-                    {
-                        $Inline['element']['non_nestables'][] = $non_nestable;
-                    }
+                    $Inline['element']['non_nestables'][] = $non_nestable;
```
aidantwoods: @PhrozenByte I did consider that, though there is no guarantee that the `$Inline['element']['non_nestables']` will exist – so the code suffers a little for readability with that added check IMO. Though I appreciate that's a subjective decision, so I'm happy to favour `array_merge` if you disagree.

Parsedown.php
aidantwoods: That sounds sensible, I'll commit that change when I get a sec :)

aidantwoods:
@PhrozenByte I think that's everything in the latest commit?

PhrozenByte:
Yes, perfect :+1:

aidantwoods:
@erusev guessing as before, you'll want camel casing instead of '`_`' as the word delimiter? 
Anything else that needs adjusting?

erusev:
@aidantwoods 

Yes, pls.

`->line('abc *abc*', 'em')` would produce `abc abc` and not parse the emphasis, right?

aidantwoods:
@erusev almost – it would be `->line('abc *abc*', array('em'))` or `->line('abc *abc*', ['em'])`, which would produce `abc *abc*` with no emphasis. 

This is because it's built to allow multiple types to be disallowed, (including where multiple non nestables are aggregated over nested elements with different non nestable conditions). So `line` does have to be able to accept an array for that to function.

erusev:
@aidantwoods 

Right, well, then there might be a better name for the parameter.

Non nestables seems to scary of a word for a parameter that just lists the types that the method should skip.

aidantwoods:
@erusev perhaps, any suggestions? 'skipInlineTypes', 'ignoredInlineTypes'?

erusev:
Something like this.

Let me have a better look at the code later next week and I might come up with some ideas.

hatrena:
Any chance to merge this PR ?

erusev:
@hatrena Sure, I'll try to review it by the end of the month.

PR #440, opened 2016-10-12
Test Parsedown against cached expect-to-pass CommonMark spec examples

PhrozenByte:
This test suite runs tests the same way as `test/CommonMarkTestWeak.php`, but uses a cached set of CommonMark spec examples in `test/commonmark/`. It is executed along with Parsedown's default test suite and runs various CommonMark spec examples, which are expected to pass. If they don't pass, the Parsedown build fails. The intention of this test suite is to make sure, that previously passed CommonMark spec examples don't fail due to unwanted side-effects of code changes.

You can re-create the `test/commonmark/` directory by executing the PHPUnit group `update`. The test suite will then run `test/CommonMarkTestWeak.php` and create files with the Markdown source and the resulting HTML markup of all passed tests. The command to execute looks like the following:

```
$ phpunit --group update
```

---

I made this because of https://github.com/erusev/parsedown/pull/439#issuecomment-253030481: Parsedown violates so many CommonMark spec examples, that it is quite impossible to make sure that no previously passed test fails. This PR causes Travis builds to fail when this happens.

However, you'll have to keep the cached expect-to-pass tests up-to-date. When merging a PR which improves CommonMark compliance, you'll have to run `phpunit --group update` and commit new passing tests (see 37f630f for an example). We can automate this with Travis; let me know if you want to do this, it's no big deal.

---

This PR is a follow-up to #423. This means that its branch bases on the branch of #423 and both the commit list and diff of this PR also include all changes of #423. You should merge #423 before merging this PR. You can then merge this PR and the duplicate commits automatically disappear (i.e. they don't get merged twice).

Here's the diff of the commits which are not already part of #423: https://github.com/PhrozenByte/parsedown/compare/bugfix/CommonMarkTest...enhancement/RequireCommonMark. Only the changes of `phpunit.xml.dist` and the new `test/CommonMarkTest.php` (see aae9b03) are relevant, all other files have been created automatically by `phpunit --group update`.


erusev:
Thanks, hopefully, I'll be able to review this by the end of the month.

PhrozenByte:
Any updates on this @erusev?

aidantwoods:
@PhrozenByte Fancy updating this pointing at my [`enhancement/types`](https://github.com/aidantwoods/parsedown/tree/enhancement/types) branch? I reckon it probably makes sense to run it alongside (or perhaps *replacing*) the commonmark weak build stage that's been running to keep track of stuff I'm fixing and to keep a history of passing tests.
e.g. latest one: https://travis-ci.org/erusev/parsedown/jobs/484852033

I've thus far been (rather crudely) glancing at the number of failures to make sure it either stays the same or goes down when I think I've fixed something. But would be nice to know if I break something I didn't notice when that number changes :p

I'm hoping things have got at least a little bit better on the CommonMark front, since (for example) I went to quite a bit of effort to add the capability to track indentation all the way down when doing recursive block parsing :p

PhrozenByte:
> I reckon it probably makes sense to run it alongside (or perhaps replacing) the commonmark weak build stage that's been running to keep track of stuff I'm fixing and to keep a history of passing tests.

I guess you should run `CommonMarkTestStrict` by default then :smile: The reason why `CommonMarkTest` extends `CommonMarkTestWeak` instead of `CommonMarkTestStrict` is because Parsedown never really cared about whitespaces. If you wanna retain whitespaces in the future, we should probably remove `CommonMarkTestWeak`.

> Fancy updating this and pointing at my [`enhancement/types`](https://github.com/aidantwoods/parsedown/tree/enhancement/types) branch?

I don't see any reason why it shouldn't work when being merged in your new branch (maybe besides switching to PSR-4?) :+1:

aidantwoods:
> I don't see any reason why it shouldn't work when being merged in your new branch (maybe besides switching to PSR-4?) 👍

Tada https://github.com/erusev/parsedown/pull/685/commits/11bc1fe1b003cf3a5678731d19845dd0c4062ca0 :D

aidantwoods:
> I guess you should run `CommonMarkTestStrict` by default then 😄 The reason why `CommonMarkTest` extends `CommonMarkTestWeak` instead of `CommonMarkTestStrict` is because Parsedown never really cared about whitespaces. If you wanna retain whitespaces in the future, we should probably remove `CommonMarkTestWeak`.

So after: 2474251, and 7513d59 there are only three tests left that passed in weak but don't pass in strict. Considerably better than the 100 or so difference that there was prior to those commits ;-)

aidantwoods:
Commits were cherry-picked onto #685

PR #441, opened 2016-10-13
Update .travis.yml

cebe:
followup to #436
- `hhvm-nightly` is not longer supported by travis: https://travis-ci.org/erusev/parsedown/jobs/167116493
- Added `nightly` which builds from PHP master. It is different from 7.1 as soon as 7.1 is released stable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: php
 
 php:
+  - nightly
   - 7.1
   - 7.0
   - 5.6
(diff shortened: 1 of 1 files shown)


PhrozenByte:
There's no need to remove `hhvm-nightly`, this incompatibility is just temporary until Travis updates its infrastructure - something that will happen sooner rather than later, Ubuntu Precise's end-of-life is in about 6 months.

gabriel-caruso:
@cebe `Travis CI` was already rewritten in #539. You can close this PR

aidantwoods:
I think already have the changes made in `master` now

Issue #443, opened 2016-10-14
Parsedown fails with long text


h2652797:
If you visit the [Parsedown Demo](http://parsedown.org/demo) and paste in this [example text](http://pastebin.com/ih1KycWK), the server fails.

The text is around 23,000 characters, and includes a single asterisk on the first line. This occurs with any text of a similar length or greater. I'm assuming Parsedown reads the opening asterisk and fails when it needs to search for the closing asterisk in such a long string.

I use Parsedown for user comments on my site. Comments are a standard MySQL text field which holds values greater than the above in length, and users occasionally include symbols like an asterisk in their comment, without a closing one. When this occurs, the comments page fails to load.

brolaugh:
I did a test on my local machine with Parsedown-extra version 0.7.1. This was my output.
> example*
example
example
example
example
example
...

maartensch:
It's probably due to the fact that the Apache 2.4.7 server configuration of the demo blocks POST request that are longer than [x]mb.

h2652797:
> It's probably due to the fact that the Apache 2.4.7 server configuration of the demo blocks POST request that are longer than [x]mb.

I don't think that's the case, because larger text blocks are accepted without an issue. For example, here's the original text with [4 more asterisks](http://pastebin.com/raw/CyGBBHeS) added throughout. It now works fine if you paste it into the demo.

There cannot be more than N characters between an opening and closing asterisk or the script fails. It may be an issue with Apache or server configuration (the demo fails the same way as my server), or Parsedown might need to be modified to handle these type of situations.

PhrozenByte:
The maximum input length depends on your PHP configuration, see https://secure.php.net/manual/de/pcre.configuration.php

cebe:
I have added a test in #444 and it looks like this file causes PHP to segfault in versions < 7.0 : https://travis-ci.org/erusev/parsedown/builds/169156782

cebe:
See also https://bugs.php.net/bug.php?id=45735

cebe:
it is likely to be caused by the [regex for emph and strong matching](https://github.com/erusev/parsedown/blob/cbc4b3f6126e484acc033ec4edf3886f319a0c0f/Parsedown.php#L1521-L1529) resulting in something simlar as described here: http://www.regular-expressions.info/catastrophic.html

cebe:
I have submitted a fix in #444

h2652797:
[...]

PR #444, opened 2016-10-20
added test file that crashes PHP

cebe:
issue #443

seem to fail on some PHP version, but not on PHP 7 for me locally.
Sending this PR to see what travis says.


cebe:
I have submitted a fix, which adds a sanity check before applying the regex, i.e. it checks whether the end marker for strong/em actually exists in the text and only then applies the regex to find it.

test/data/dos.md
```diff
@@ -0,0 +1,2549 @@
+Without additional checks, this file causes a segfault in PHP < 7.0, which is caused by Regex Backtrack*
```
aidantwoods: Could we maybe generate these results at runtime in some dedicated tests, as opposed to adding 10k lines to the repo? :p

aidantwoods (changes requested):
Thanks for adding a sanity check along with the test cases :)

test/data/dos.md
cebe: given that all tests work in the same way by reading files imo its the most straight forward way to just have these files here. isn't disk space is cheap enough? :)

aidantwoods:
Disk space may be cheap, but not sure about network speed – cloning the
repo will take quite a bit longer :p
Also IMO reading a script that generates this makes the intent and
structure of the test case more clear (as opposed to scrolling through a
huge file)

Also we have a few exceptions to reading files in test cases now, so
writing tests in code won't be without precedent.

cebe:
feel free to fix this the way you like it more. I currently don't have time to work on this.

jcapinc:
So we want the really large example files removed? I am looking for a way to pitch in, can I clone and apply this pull request and submit a new one to fix this issue?

cebe:
feel free to cherry-pick my commits to a new branch/PR.

PR #447, opened 2016-11-01
Fix include from ParsedownTest

greut:
I wasn't able to run all the tests from ParsedownExtra because of it.

### DIFF
--- a/test/ParsedownTest.php
+++ b/test/ParsedownTest.php
@@ -139,7 +139,7 @@ EXPECTED_HTML;
 
     public function testLateStaticBinding()
     {
-        include 'test/TestParsedown.php';
+        include __DIR__ . '/TestParsedown.php';
 
         $parsedown = Parsedown::instance();
         $this->assertInstanceOf('Parsedown', $parsedown);


erusev:
@hkdobrev do you have any thoughts on this?

test/ParsedownTest.php
```diff
@@ -139,7 +139,7 @@ public function test_no_markup()
 
     public function testLateStaticBinding()
     {
-        include 'test/TestParsedown.php';
+        include dirname(__file__) . '/TestParsedown.php';
```
hkdobrev: I think we could just use `__DIR__` instead of `dirname(__FILE__)` as Parsedown no longer supports PHP 5.2.

hkdobrev:
@erusev I don't have anything else to add other than the comment above. Seems legit.

@greut Could you elaborate more on what problem you have experienced when running parsedown-extra tests? Probably something has changed as it wasn't a problem before?

greut:
```
PHPUnit 5.6.2 by Sebastian Bergmann and contributors.

...........................................................E      60 / 60 (100%)

Time: 53 ms, Memory: 6.00MB

There was 1 error:

1) ParsedownExtraTest::testLateStaticBinding
include(test/TestParsedown.php): failed to open stream: No such file or directory

/srv/http/pouille2/vendor/erusev/parsedown/test/ParsedownTest.php:142
/srv/http/pouille2/vendor/erusev/parsedown/test/ParsedownTest.php:142

ERRORS!
Tests: 60, Assertions: 59, Errors: 1.
```

greut:
The problem occurs when I'm running the parsedown-extra tests from the parsedown-extra directory in the case where `parsedown` is the parent (sibling).
https://github.com/erusev/parsedown-extra/blob/master/test/bootstrap.php#L3-L5

```
erusev/parsedown
erusev/parsedown-extra
```

greut:
Thanks!

PR #452, opened 2016-12-08
Add a function to read remote file

FrederickChan:
Implementation 
-------------------
````
$text = new Parsedown();
$filepath = dirname(__FILE__).'/example.md
echo $text->read_file( $filepath);
````

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -21,6 +21,21 @@ class Parsedown
 
     # ~
 
+     /**
+     * Reads a file and returns the parsed markdown text.
+     * @param $file_path
+     * @return string
+     */
+    function read_file($file_path) {
+        if (file_exists($file_path)) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: Parsedown.php [hunk 1]
```diff
@@ -21,6 +21,21 @@ class Parsedown
 
     # ~
 
+     /**
+     * Reads a file and returns the parsed markdown text.
+     * @param $file_path
+     * @return string
+     */
+    function read_file($file_path) {
+        if (file_exists($file_path)) {
```
cebe: non-existing file should error instead of returning empty string.

FrederickChan: Hi, the reason I didn't put down an error is because of locale issues.

aidantwoods:
Since it's only one line to do this manually – it might be best to leave this up to the user so we don't make assumptions about error handling on their behalf?

PR #459, opened 2017-01-07
fix_inlineLink_regex

### LINKED ISSUE #458: preg_match in inlineLink() causes segfault
the preg_match pattern
``/^[(]((?:[^ ()]|[(][^ )]+[)])+)(?:[ ]+("[^"]*"|\'[^\']*\'))?[)]/``
causes a segfault on a huge link or an attempt to insert a base64-image

The parts in the alternation group are mutually exclusive, so it should be no problem to an additonal quantifier (+ or ++) for first alternative and a possessive quantifier for the group.
``/^[(]((?:[^ ()]++|[(][^ )]+[)])++)(?:[ ]+("[^"]*"|\'[^\']*\'))?[)]/``

gene-sis:
adjust two regex pattern within inlineLink() to reduce backtracking
add test with base64 image
fixes #458


gene-sis:
the test test_no_markup() returns an error in my local environment, but this error already occurred before making any changes.

all test_() assertions would pass with 
`ini_set("pcre.recursion_limit","31");'

Testing the base64 image with the old regexs will pass with
`ini_set("pcre.recursion_limit","11283");`
The reason for that should be, that phpunit uses a higher stack size and therefore allow a higher recursion limit

erusev:
Looks great, thanks!

Issue #265, opened 2015-01-14
Broken Headlines in the latest version


bastianallgeier:
Unfortunately the latest version comes with this pretty massive bug: 

```
##heading
```
is being converted to

```
<p>##heading</p>
```

erusev:
@bastianallgeier 

This is expected. Headers need a space between the `#` and the text. This is dictated by the CommonMark spec. You are the second person to complain about it. I should perhaps change it back.

rhukster:
Even in the **not-so-specific** Gruber spec, they have spaces:

http://daringfireball.net/projects/markdown/syntax#header

erusev:
@rhukster 

True, but it's likely that a huge number of Markdown texts don't have these spaces.

bastianallgeier:
I didn't even find it myself. I added it to Kirby's develop branch a couple hours ago and already got a ticket for it. I get that it's officially not documented that way, but I'm sure that lots of people write it like that.

erusev:
@bastianallgeier 

I believe that to be true. I'll release a patch within the next 10 min.

rhukster:
True, I guess it depends on how close to the CommonMark spec you are trying to get.  Is it possible to have a toggle/option at some point to enforce markdown or legacy?

bastianallgeier:
@erusev awesome! I think it's a lot about usability in this case, so I'm glad you see it this way as well.

bastianallgeier:
@rhukster @erusev I wouldn't add too many "toggles". They add complexity and probably negatively affect the performance. Where's the harm if it does follow the specs but also is slightly more forgiving?

erusev:
The release is out.

erusev:
@rhukster 

It is possible. It could be just one toggle. I'm considering it.

cebe:
The only conflict this has is that if you write markdown like this:

```

- #123 abc
- #456 def

```

you will get headlines instead of issue numbers...

rhukster:
Well in this case the CommonMark explicitly says:

> The opening sequence of # characters cannot be followed directly by a nonspace character.

http://spec.commonmark.org/0.14/#atx-headers

If the spec didn't mention it, then it would be fine.  But the whole point of CommonMark is to tidy up issues related to lack of detail in the original spec. 

I know changes like this could potentially break lots of people's content, hence the idea for the toggle.  This forgiving change actually makes Parsedown less CommonMark compatible.

rhukster:
[...]

PR #464, opened 2017-01-18
Fix #463

### LINKED ISSUE #463: Headers should be a serie of '#' followed by at least one whitespace
Using Parsedown 1.6.

Currently, this

     #nottitle 

will render as:

    <h1>nottitle</h1>

As specified in [CommonMark specification](http://spec.commonmark.org/0.26/#atx-headings):

> The opening sequence of # characters must be followed by a space or by the end of line.

Also:

> At least one space is required between the # characters and the heading’s contents, unless the heading is empty. Note that many implementations currently do not require the space. However, the space was required by the original ATX implementation

Expected result, same as Github:

#nottitle 

Use case: I extended Parsedown to support `#hashtags` in my software, and hashtags set at the beginning of a new line now generate a header.

EDIT: that's probably not the best way to do it, but I added this to my Parsedown extension to fix it:

```php
protected function blockHeader($line)
{
    if (preg_match('/^#+ /', $line['text'])) {
        return parent::blockHeader($line);
    }
}
```

aidantwoods:


erusev:
Please, see https://github.com/erusev/parsedown/issues/265

aidantwoods:
Ah I see – guess that setter for it was never added?

aidantwoods:
Closing in favour of #589

PR #465, opened 2017-01-21
Trim surrounding whitespace from URL in inlineLink

aidantwoods:
Fixes https://github.com/erusev/parsedown-extra/issues/103

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1217,7 +1217,7 @@ class Parsedown
             return;
         }
 
-        if (preg_match('/^[(]((?:[^ ()]++|[(][^ )]+[)])++)(?:[ ]+("[^"]*"|\'[^\']*\'))?[)]/', $remainder, $matches))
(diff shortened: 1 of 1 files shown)


kminchev:
Why this PR is in this repo?

aidantwoods:
@kminchev 

https://github.com/erusev/parsedown-extra/pull/104#issuecomment-274251222

Also apparently Commonmark reference parser produces the expected result from that issue – so this is really a bug in Parsedown 😉

Parsedown.php
```diff
@@ -1217,7 +1217,7 @@ protected function inlineLink($Excerpt)
             return;
         }
 
-        if (preg_match('/^[(]((?:[^ ()]++|[(][^ )]+[)])++)(?:[ ]+("[^"]*"|\'[^\']*\'))?[)]/', $remainder, $matches))
+        if (preg_match('/^[(]\s*+((?:[^ ()]++|[(][^ )]+[)])++)(?:[ ]+("[^"]*"|\'[^\']*\'))?\s*[)]/', $remainder, $matches))
```
erusev: what's the role of the `+` in 
image

aidantwoods: Possessive quantifier – probably not necessary here actually, I tend to use them by habit if I know backtracking won't be needed

PR #491, opened 2017-04-10
Fixed README Caret link

TonyRouse:
Might be the site that should be fixed but passing the ?ref tag seems to not allow the site to be visited, removing it worked for me.


README.md
```diff
@@ -1,4 +1,4 @@
-> You might also like [Caret](http://caret.io?ref=parsedown) - our Markdown editor for Mac / Windows / Linux.
+> You might also like [Caret](http://caret.io) - our Markdown editor for Mac / Windows / Linux.
```
aidantwoods: Instead of removing the `ref` parameter, could you instead add the `/` that should be there. Also, since there's a redirect in place anyway, we might as well link to the https version too :)

i.e. instead change it to `https://caret.io/?ref=parsedown`

PR #495, opened 2017-05-01
Prevent various XSS attacks [rebase and update of #276]

### LINKED ISSUE #161: Safe Mode

### LINKED ISSUE #497: Link allows javascript which could lead to XSS issues
So I don't know if this is a bug or an expected feature, so please correct me if I'm wrong, but I found that I actually can write javascript links like this:

```
[test](javascript:alert%281%29)
```

An alert does pop out and this behavior is actually disallowed in GitHub. 

People can do nasty things if a victim does click on one of these links.

### LINKED ISSUE #530: Cross-site scripting vulnerability in ParseDown
```php
<?php

include("parsedown/Parsedown.php");

echo Parsedown::instance()->text('Hello ![google" onerror="alert(1)](http://google.com)'); 
````

The code above simply creates an instance of your Parsedown parser and then creates an image in markdown. This creation of the image is vulnerable to XSS. I would advice something like HTML encoding some of the input in certain places (especially for things like quotes).

### PR DESCRIPTION
**Pull request by aidantwoods**
I've picked up on the work started over at https://github.com/erusev/parsedown/pull/276 and rebased on `erusev/master`.

Since this is rebased on master, I can't point at PR at `naNuke/master` without running into the merge conflicts that I've already resolved manually.

I've implemented what I suggested [earlier](https://github.com/erusev/parsedown/pull/276#issuecomment-298244798) so that all attributes are properly encoded (and not just the specific ones we remember).

I've also added some tests, so @erusev's concern here should hopefully now be resolved, albeit a year later 😉 
https://github.com/erusev/parsedown/pull/276#issuecomment-178577968
> @malas one reason is the lack of tests

~~One thing to note is that all this can be circumvented if you forget to turn on `$Parsedown->setMarkupEscaped(true);`
(which is off by default) as you could just write a script tag manually for xss (even though the attributes and link destinations will be safe). So let's all remember to enable this setting 😉~~

Attributes are now always escaped properly (this speaks to just outputting things correctly), but link based XSS or XSS from writing plain old script tags will only be prevented only if the new `setSafeMode` is enabled.

```php
$Parsedown->setSafeMode(true);
```

---

Closes #161
Closes #497
Closes #276
Closes #403 
Closes #530

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1513,6 +1513,22 @@ class Parsedown
             $Element = $this->filterUnsafeUrlInAttribute($Element, $safeUrlNameToAtt[$Element['name']]);
         }
 
+        if ( ! empty($Element['attributes']))
+        {
+            # clear out nulls
+            $Element['attributes'] = array_filter(
+                $Element['attributes'],
+                function ($v) {return $v !== null;}
+            );
+
+            $onEventAttributes = preg_grep('/^\s*+on/i', array_flip($Element['attributes']));
+
+            foreach ($onEventAttributes as $att)
+            {
+                unset($Element['attributes'][$att]);
+            }
+        }
+
         return $Element;
     }
 
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1533,7 +1533,7 @@ class Parsedown
                     unset($Element['attributes'][$att]);
                 }
                 # dump onevent attribute
-                elseif (preg_match('/^on/i', $att))
+                elseif (self::striAtStart($att, 'on'))
                 {
                     unset($Element['attributes'][$att]);
                 }
@@ -1551,7 +1551,7 @@ class Parsedown
 
             foreach ($this->safeLinksWhitelist as $scheme)
             {
-                if (stripos($Element['attributes'][$attribute], $scheme) === 0)
+                if (self::striAtStart($Element['attributes'][$attribute], $scheme))
                 {
                     $safe = true;
 
@@ -1582,6 +1582,20 @@ class Parsedown
     protected static function escape($text, $allowQuotes = false)
     {
         return htmlspecialchars($text, $allowQuotes ? ENT_NOQUOTES : ENT_QUOTES, 'UTF-8');
+    }
+
+    protected static function striAtStart($string, $needle)
+    {
+        $len = strlen($needle);
+
+        if ($len > strlen($string))
+        {
+            return false;
+        }
+        else
+        {
+            return strtolower(substr($string, 0, $len)) === strtolower($needle);
+        }
     }
 
     static function instance($name = 'default')
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1503,7 +1503,8 @@ class Parsedown
 
     protected function sanitiseElement(array $Element)
     {
-        $safeUrlNameToAtt = array(
+        static $badAttributeChars = "\"'= \t\n\r\0\x0B";
+        static $safeUrlNameToAtt  = array(
             'a'   => 'href',
             'img' => 'src',
         );
@@ -1515,13 +1516,21 @@ class Parsedown
 
         if ( ! empty($Element['attributes']))
         {
-            # clear out nulls
-            $Element['attributes'] = array_filter(
-                $Element['attributes'],
-                function ($v) {return $v !== null;}
-            );
+            foreach ($Element['attributes'] as $att => $val)
+            {
+                # clear out nulls
+                if ($val === null)
+                {
+                    unset($Element['attributes'][$att]);
+                }
+                # filter out badly parsed attribute
+                elseif (strpbrk($att, $badAttributeChars) !== false)
+                {
+                    unset($Element['attributes'][$att]);
+                }
+            }
 
-            $onEventAttributes = preg_grep('/^\s*+on/i', array_flip($Element['attributes']));
+            $onEventAttributes = preg_grep('/^on/i', array_flip($Element['attributes']));
 
             foreach ($onEventAttributes as $att)
             {
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1530,9 +1530,9 @@ class Parsedown
                 }
             }
 
-            $onEventAttributes = preg_grep('/^on/i', array_flip($Element['attributes']));
+            $onEventAttributeKeys = preg_grep('/^on/i', array_keys($Element['attributes']));
 
-            foreach ($onEventAttributes as $att)
+            foreach ($onEventAttributeKeys as $att)
             {
                 unset($Element['attributes'][$att]);
             }
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1262,13 +1262,18 @@ class Parsedown
             $Element['attributes']['title'] = $Definition['title'];
         }
 
-        if ( $this->safeLinksEnabled && stripos($Element['attributes']['href'], 'javascript:') === 0 )
+        if ( $this->safeLinksEnabled && preg_match("/^(\/|https?:\/\/|ftps?:\/\/)/ui", $Element['attributes']['href']) === 0 )
         {
             return;
         }
 
-        $Element['attributes']['href'] = htmlspecialchars($Element['attributes']['href']);
-        $Element['text'] = htmlspecialchars($Element['text']);
+        $Element['attributes']['href'] = htmlspecialchars($Element['attributes']['href'], ENT_QUOTES);
+        $Element['text'] = htmlspecialchars($Element['text'], ENT_QUOTES);
+
+        if ( $Element['attributes']['title'] !== null )
+        {
+            $Element['attributes']['title'] = htmlspecialchars($Element['attributes']['title'], ENT_QUOTES);
+        }
 
         return array(
             'extent' => $extent,
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -97,8 +97,6 @@ class Parsedown
         'ircs:',
         'git:',
         'ssh:',
-        'ftp:',
-        'ftps:',
         'news:',
         'steam:',
     );
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -91,7 +91,9 @@ class Parsedown
         'ftp://',
         'ftps://',
         'mailto:',
-        'data:image/png;',
+        'data:image/png;base64,',
+        'data:image/gif;base64,',
+        'data:image/jpg;base64,',
     );
 
     #
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1554,14 +1554,7 @@ class Parsedown
             }
         }
 
-        $Element['attributes'][$attribute] = preg_replace_callback(
-            '/[^\/#?&=%]++/',
-            function (array $match)
-            {
-                return urlencode($match[0]);
-            },
-            $Element['attributes'][$attribute]
-        );
+        $Element['attributes'][$attribute] = str_replace(':', '%3A', $Element['attributes'][$attribute]);
 
         return $Element;
     }
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -89,7 +89,9 @@ class Parsedown
         'https://',
         '/',
         'ftp://',
-        'ftps://'
+        'ftps://',
+        'mailto:',
+        'data:image/png;',
     );
 
     #
@@ -359,8 +361,6 @@ class Parsedown
     {
         $text = $Block['element']['text']['text'];
 
-        $text = htmlspecialchars($text, ENT_NOQUOTES, 'UTF-8');
-
         $Block['element']['text']['text'] = $text;
 
         return $Block;
@@ -422,7 +422,7 @@ class Parsedown
 
             if (isset($matches[1]))
             {
-                $class = 'language-'.htmlspecialchars($matches[1], ENT_QUOTES, 'UTF-8');
+                $class = 'language-'.$matches[1];
 
                 $Element['attributes'] = array(
                     'class' => $class,
@@ -473,8 +473,6 @@ class Parsedown
     protected function blockFencedCodeComplete($Block)
     {
         $text = $Block['element']['text']['text'];
-
-        $text = htmlspecialchars($text, ENT_NOQUOTES, 'UTF-8');
 
         $Block['element']['text']['text'] = $text;
 
@@ -1091,7 +1089,6 @@ class Parsedown
         if (preg_match('/^('.$marker.'+)[ ]*(.+?)[ ]*(?<!'.$marker.')\1(?!'.$marker.')/s', $Excerpt['text'], $matches))
         {
             $text = $matches[2];
-            $text = htmlspecialchars($text, ENT_NOQUOTES, 'UTF-8');
             $text = preg_replace("/[ ]*\n/", ' ', $text);
 
             return array(
@@ -1108,7 +1105,7 @@ class Parsedown
     {
         if (strpos($Excerpt['text'], '>') !== false and preg_match('/^<((mailto:)?\S+?@\S+?)>/i', $Excerpt['text'], $matches))
         {
-            $url = htmlspecialchars($matches[1], ENT_QUOTES, 'UTF-8');
+            $url = $matches[1];
 
             if ( ! isset($matches[2]))
             {
@@ -1270,32 +1267,6 @@ class Parsedown
             $Element['attributes']['title'] = $Definition['title'];
         }
 
-        if ( $this->safeLinksEnabled )
-        {
-            $matched = false;
-            foreach ( $this->safeLinksWhitelist as $scheme )
-            {
-                if ( stripos($Element['attributes']['href'], $scheme) === 0 )
-                {
-                    $matched = true;
-                    break;
-                }
-            }
-
-            if ( ! $matched )
-            {
-                return;
-            }
-        }
-
-        $Element['attributes']['href'] = htmlspecialchars($Element['attributes']['href'], ENT_QUOTES, 'UTF-8');
-        $Element['text'] = htmlspecialchars($Element['text'], ENT_QUOTES, 'UTF-8');
-
-        if ( $Element['attributes']['title'] !== null )
-        {
-            $Element['attributes']['title'] = htmlspecialchars($Element['attributes']['title'], ENT_QUOTES, 'UTF-8');
-        }
-
         return array(
             'extent' => $extent,
             'element' => $Element,
@@ -1384,7 +1355,7 @@ class Parsedown
 
         if (preg_match('/\bhttps?:[\/]{2}[^\s<]+\b\/*/ui', $Excerpt['context'], $matches, PREG_OFFSET_CAPTURE))
         {
-            $url = htmlspecialchars($matches[0][0], ENT_QUOTES, 'UTF-8');
+            $url = $matches[0][0];
 
             $Inline = array(
                 'extent' => strlen($matches[0][0]),
@@ -1406,7 +1377,7 @@ class Parsedown
     {
         if (strpos($Excerpt['text'], '>') !== false and preg_match('/^<(\w+:\/{2}[^ >]+)>/i', $Excerpt['text'], $matches))
         {
-            $url = htmlspecialchars($matches[1], ENT_QUOTES, 'UTF-8');
+            $url = $matches[1];
 
             return array(
                 'extent' => strlen($matches[0]),
@@ -1444,6 +1415,8 @@ class Parsedown
 
     protected function element(array $Element)
     {
+        $Element = $this->sanitiseElement($Element);
+
         $markup = '<'.$Element['name'];
 
         if (isset($Element['attributes']))
@@ -1455,7 +1428,7 @@ class Parsedown
                     continue;
                 }
 
-                $markup .= ' '.$name.'="'.$value.'"';
+                $markup .= ' '.$name.'="'.self::escape($value).'"';
             }
         }
 
@@ -1469,7 +1442,7 @@ class Parsedown
             }
             else
             {
-                $markup .= $Element['text'];
+                $markup .= self::escape($Element['text'], true);
             }
 
             $markup .= '</'.$Element['name'].'>';
@@ -1528,9 +1501,54 @@ class Parsedown
         return $markup;
     }
 
+    protected function sanitiseElement(array $Element)
+    {
+        $safeUrlNameToAtt = array(
+            'a'   => 'href',
+            'img' => 'src',
+        );
+
+        if (isset($safeUrlNameToAtt[$Element['name']]))
+        {
+            $Element = $this->filterUnsafeUrlInAttribute($Element, $safeUrlNameToAtt[$Element['name']]);
+        }
+
+        return $Element;
+    }
+
+    protected function filterUnsafeUrlInAttribute(array $Element, $attribute)
+    {
+        if ($this->safeLinksEnabled)
+        {
+            $safe = false;
+
+            foreach ($this->safeLinksWhitelist as $scheme)
+            {
+                if (stripos($Element['attributes'][$attribute], $scheme) === 0)
+                {
+                    $safe = true;
+
+                    break;
+                }
+            }
+
+            if ( ! $safe)
+            {
+                unset($Element['attributes'][$attribute]);
+            }
+        }
+
+        return $Element;
+    }
+
     #
     # Static Methods
     #
+
+    protected static function escape($text, $allowQuotes = false)
+    {
+        return htmlspecialchars($text, $allowQuotes ? ENT_NOQUOTES : ENT_QUOTES, 'UTF-8');
+    }
 
     static function instance($name = 'default')
     {
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1422,7 +1422,10 @@ class Parsedown
 
     protected function element(array $Element)
     {
-        $Element = $this->sanitiseElement($Element);
+        if ($this->safeMode)
+        {
+            $Element = $this->sanitiseElement($Element);
+        }
 
         $markup = '<'.$Element['name'];
 
@@ -1543,26 +1546,22 @@ class Parsedown
 
     protected function filterUnsafeUrlInAttribute(array $Element, $attribute)
     {
-        if ($this->safeMode)
+        foreach ($this->safeLinksWhitelist as $scheme)
         {
-            foreach ($this->safeLinksWhitelist as $scheme)
+            if (self::striAtStart($Element['attributes'][$attribute], $scheme))
             {
-                if (self::striAtStart($Element['attributes'][$attribute], $scheme))
-                {
-                    return $Element;
-                }
+                return $Element;
             }
-
-            $Element['attributes'][$attribute] = preg_replace_callback(
-                '/[^\/#?&=%]++/',
-                function (array $match)
-                {
-                    return urlencode($match[0]);
-                },
-                $Element['attributes'][$attribute]
-            );
-
         }
+
+        $Element['attributes'][$attribute] = preg_replace_callback(
+            '/[^\/#?&=%]++/',
+            function (array $match)
+            {
+                return urlencode($match[0]);
+            },
+            $Element['attributes'][$attribute]
+        );
 
         return $Element;
     }
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -422,7 +422,7 @@ class Parsedown
 
             if (isset($matches[1]))
             {
-                $class = 'language-'.$matches[1];
+                $class = 'language-'.htmlspecialchars($matches[1], ENT_QUOTES, 'UTF-8');
 
                 $Element['attributes'] = array(
                     'class' => $class,
@@ -532,10 +532,10 @@ class Parsedown
                 ),
             );
 
-            if($name === 'ol') 
+            if($name === 'ol')
             {
                 $listStart = stristr($matches[0], '.', true);
-                
+
                 if($listStart !== '1')
                 {
                     $Block['element']['attributes'] = array('start' => $listStart);
@@ -1108,7 +1108,7 @@ class Parsedown
     {
         if (strpos($Excerpt['text'], '>') !== false and preg_match('/^<((mailto:)?\S+?@\S+?)>/i', $Excerpt['text'], $matches))
         {
-            $url = $matches[1];
+            $url = htmlspecialchars($matches[1], ENT_QUOTES, 'UTF-8');
 
             if ( ! isset($matches[2]))
             {
@@ -1288,12 +1288,12 @@ class Parsedown
             }
         }
 
-        $Element['attributes']['href'] = htmlspecialchars($Element['attributes']['href'], ENT_QUOTES);
-        $Element['text'] = htmlspecialchars($Element['text'], ENT_QUOTES);
+        $Element['attributes']['href'] = htmlspecialchars($Element['attributes']['href'], ENT_QUOTES, 'UTF-8');
+        $Element['text'] = htmlspecialchars($Element['text'], ENT_QUOTES, 'UTF-8');
 
         if ( $Element['attributes']['title'] !== null )
         {
-            $Element['attributes']['title'] = htmlspecialchars($Element['attributes']['title'], ENT_QUOTES);
+            $Element['attributes']['title'] = htmlspecialchars($Element['attributes']['title'], ENT_QUOTES, 'UTF-8');
         }
 
         return array(
@@ -1384,14 +1384,16 @@ class Parsedown
 
         if (preg_match('/\bhttps?:[\/]{2}[^\s<]+\b\/*/ui', $Excerpt['context'], $matches, PREG_OFFSET_CAPTURE))
         {
+            $url = htmlspecialchars($matches[0][0], ENT_QUOTES, 'UTF-8');
+
             $Inline = array(
                 'extent' => strlen($matches[0][0]),
                 'position' => $matches[0][1],
                 'element' => array(
                     'name' => 'a',
-                    'text' => $matches[0][0],
+                    'text' => $url,
                     'attributes' => array(
-                        'href' => $matches[0][0],
+                        'href' => $url,
                     ),
                 ),
             );
@@ -1404,7 +1406,7 @@ class Parsedown
     {
         if (strpos($Excerpt['text'], '>') !== false and preg_match('/^<(\w+:\/{2}[^ >]+)>/i', $Excerpt['text'], $matches))
         {
-            $url = str_replace(array('&', '<'), array('&amp;', '&lt;'), $matches[1]);
+            $url = htmlspecialchars($matches[1], ENT_QUOTES, 'UTF-8');
 
             return array(
                 'extent' => strlen($matches[0]),
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -93,6 +93,14 @@ class Parsedown
         'data:image/png;base64,',
         'data:image/gif;base64,',
         'data:image/jpeg;base64,',
+        'irc:',
+        'ircs:',
+        'git:',
+        'ssh:',
+        'ftp:',
+        'ftps:',
+        'news:',
+        'steam:',
     );
 
     #
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -93,7 +93,7 @@ class Parsedown
         'mailto:',
         'data:image/png;base64,',
         'data:image/gif;base64,',
-        'data:image/jpg;base64,',
+        'data:image/jpeg;base64,',
     );
 
     #
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -75,6 +75,15 @@ class Parsedown
 
     protected $urlsLinked = true;
 
+    function setSafeLinksEnabled($safeLinksEnabled)
+    {
+        $this->safeLinksEnabled = $safeLinksEnabled;
+
+        return $this;
+    }
+
+    protected $safeLinksEnabled = true;
+
     #
     # Lines
     #
@@ -1253,7 +1262,13 @@ class Parsedown
             $Element['attributes']['title'] = $Definition['title'];
         }
 
-        $Element['attributes']['href'] = str_replace(array('&', '<'), array('&amp;', '&lt;'), $Element['attributes']['href']);
+        if ( $this->safeLinksEnabled && stripos($Element['attributes']['href'], 'javascript:') === 0 )
+        {
+            return;
+        }
+
+        $Element['attributes']['href'] = htmlspecialchars($Element['attributes']['href']);
+        $Element['text'] = htmlspecialchars($Element['text']);
 
         return array(
             'extent' => $extent,
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1505,7 +1505,7 @@ class Parsedown
 
     protected function sanitiseElement(array $Element)
     {
-        static $badAttributeChars = "\"'= \t\n\r\0\x0B";
+        static $goodAttribute = '/^[a-zA-Z0-9][a-zA-Z0-9-_]*+$/';
         static $safeUrlNameToAtt  = array(
             'a'   => 'href',
             'img' => 'src',
@@ -1520,23 +1520,16 @@ class Parsedown
         {
             foreach ($Element['attributes'] as $att => $val)
             {
-                # clear out nulls
-                if ($val === null)
-                {
-                    unset($Element['attributes'][$att]);
-                }
                 # filter out badly parsed attribute
-                elseif (strpbrk($att, $badAttributeChars) !== false)
+                if ( ! preg_match($goodAttribute, $att))
                 {
                     unset($Element['attributes'][$att]);
                 }
-            }
-
-            $onEventAttributeKeys = preg_grep('/^on/i', array_keys($Element['attributes']));
-
-            foreach ($onEventAttributeKeys as $att)
-            {
-                unset($Element['attributes'][$att]);
+                # dump onevent attribute
+                elseif (preg_match('/^on/i', $att))
+                {
+                    unset($Element['attributes'][$att]);
+                }
             }
         }
 
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1547,29 +1547,23 @@ class Parsedown
     {
         if ($this->safeLinksEnabled)
         {
-            $safe = false;
-
             foreach ($this->safeLinksWhitelist as $scheme)
             {
                 if (self::striAtStart($Element['attributes'][$attribute], $scheme))
                 {
-                    $safe = true;
-
-                    break;
+                    return $Element;
                 }
             }
 
-            if ( ! $safe)
-            {
-                $Element['attributes'][$attribute] = preg_replace_callback(
-                    '/[^\/#?&=%]++/',
-                    function (array $match)
-                    {
-                        return urlencode($match[0]);
-                    },
-                    $Element['attributes'][$attribute]
-                );
-            }
+            $Element['attributes'][$attribute] = preg_replace_callback(
+                '/[^\/#?&=%]++/',
+                function (array $match)
+                {
+                    return urlencode($match[0]);
+                },
+                $Element['attributes'][$attribute]
+            );
+
         }
 
         return $Element;
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -87,7 +87,6 @@ class Parsedown
     protected $safeLinksWhitelist = array(
         'http://',
         'https://',
-        '/',
         'ftp://',
         'ftps://',
         'mailto:',
@@ -1554,7 +1553,14 @@ class Parsedown
 
             if ( ! $safe)
             {
-                unset($Element['attributes'][$attribute]);
+                $Element['attributes'][$attribute] = preg_replace_callback(
+                    '/[^\/#?&=%]++/',
+                    function (array $match)
+                    {
+                        return urlencode($match[0]);
+                    },
+                    $Element['attributes'][$attribute]
+                );
             }
         }
 
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -84,6 +84,14 @@ class Parsedown
 
     protected $safeLinksEnabled = true;
 
+    protected $safeLinksWhitelist = array(
+        'http://',
+        'https://',
+        '/',
+        'ftp://',
+        'ftps://'
+    );
+
     #
     # Lines
     #
@@ -1262,9 +1270,22 @@ class Parsedown
             $Element['attributes']['title'] = $Definition['title'];
         }
 
-        if ( $this->safeLinksEnabled && preg_match("/^(\/|https?:\/\/|ftps?:\/\/)/ui", $Element['attributes']['href']) === 0 )
+        if ( $this->safeLinksEnabled )
         {
-            return;
+            $matched = false;
+            foreach ( $this->safeLinksWhitelist as $scheme )
+            {
+                if ( stripos($Element['attributes']['href'], $scheme) === 0 )
+                {
+                    $matched = true;
+                    break;
+                }
+            }
+
+            if ( ! $matched )
+            {
+                return;
+            }
         }
 
         $Element['attributes']['href'] = htmlspecialchars($Element['attributes']['href'], ENT_QUOTES);
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -75,14 +75,14 @@ class Parsedown
 
     protected $urlsLinked = true;
 
-    function setSafeLinksEnabled($safeLinksEnabled)
+    function setSafeMode($safeMode)
     {
-        $this->safeLinksEnabled = $safeLinksEnabled;
+        $this->safeMode = (bool) $safeMode;
 
         return $this;
     }
 
-    protected $safeLinksEnabled = true;
+    protected $safeMode;
 
     protected $safeLinksWhitelist = array(
         'http://',
@@ -378,7 +378,7 @@ class Parsedown
 
     protected function blockComment($Line)
     {
-        if ($this->markupEscaped)
+        if ($this->markupEscaped or $this->safeMode)
         {
             return;
         }
@@ -700,7 +700,7 @@ class Parsedown
 
     protected function blockMarkup($Line)
     {
-        if ($this->markupEscaped)
+        if ($this->markupEscaped or $this->safeMode)
         {
             return;
         }
@@ -1282,7 +1282,7 @@ class Parsedown
 
     protected function inlineMarkup($Excerpt)
     {
-        if ($this->markupEscaped or strpos($Excerpt['text'], '>') === false)
+        if ($this->markupEscaped or $this->safeMode or strpos($Excerpt['text'], '>') === false)
         {
             return;
         }
@@ -1543,7 +1543,7 @@ class Parsedown
 
     protected function filterUnsafeUrlInAttribute(array $Element, $attribute)
     {
-        if ($this->safeLinksEnabled)
+        if ($this->safeMode)
         {
             foreach ($this->safeLinksWhitelist as $scheme)
             {
--- a/test/ParsedownTest.php
+++ b/test/ParsedownTest.php
@@ -46,7 +46,7 @@ class ParsedownTest extends PHPUnit_Framework_TestCase
         $expectedMarkup = str_replace("\r\n", "\n", $expectedMarkup);
         $expectedMarkup = str_replace("\r", "\n", $expectedMarkup);
 
-        $this->Parsedown->setMarkupEscaped($test === 'xss_text_encoding');
+        $this->Parsedown->setSafeMode(substr($test, 0, 3) === 'xss');
 
         $actualMarkup = $this->Parsedown->text($markdown);
 
--- a/test/ParsedownTest.php
+++ b/test/ParsedownTest.php
@@ -46,6 +46,8 @@ class ParsedownTest extends PHPUnit_Framework_TestCase
         $expectedMarkup = str_replace("\r\n", "\n", $expectedMarkup);
         $expectedMarkup = str_replace("\r", "\n", $expectedMarkup);
 
+        $this->Parsedown->setMarkupEscaped($test === 'xss_text_encoding');
+
         $actualMarkup = $this->Parsedown->text($markdown);
 
         $this->assertEquals($expectedMarkup, $actualMarkup);
--- a/test/data/inline_link.html
+++ b/test/data/inline_link.html
@@ -1,5 +1,5 @@
 <p><a href="http://example.com">link</a></p>
-<p><a href="/url-%28parentheses%29">link</a> with parentheses in URL </p>
+<p><a href="/url-(parentheses)">link</a> with parentheses in URL </p>
 <p>(<a href="/index.php">link</a>) in parentheses</p>
 <p><a href="http://example.com"><code>link</code></a></p>
 <p><a href="http://example.com"><img src="http://parsedown.org/md.png" alt="MD Logo" /></a></p>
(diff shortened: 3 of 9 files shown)


aidantwoods:
https://github.com/erusev/parsedown/pull/495/commits/131ba758514c5e905663bd82fc7c1c8271c7edc5 addresses a small case study.

I noticed that someone had written an extension for ParseDown extra here https://github.com/tovic/parsedown-extra-plugin

One of the things it does it expand parsedown extra's attribute block feature (i.e. `# Header {.sth}` => `<h1 class="sth">Header</h1>`).

The extension allows arbitrary attributes to be set, as well as being vulnerable itself to xss via attribute encoding.

So running this extension on current master, we see that it's vulnerable because it does it's own element parsing.
screen shot 2017-05-01 at 15 23 50

Applying everything upto https://github.com/erusev/parsedown/pull/495/commits/af04ac92e2ff852309891ebc767fa5a6bf179f39 gets us this far:
screen shot 2017-05-01 at 15 26 02

Which is great IMO! This feature is doing what it's meant to – sanitising *all* attributes (even ones Parsedown base didn't create).

But because this extension allows arbitrary attributes to be set, we can still achieve something like this:
screen shot 2017-05-01 at 15 31 28

Which is XSS without needing to break out of the quote.

So https://github.com/erusev/parsedown/pull/495/commits/131ba758514c5e905663bd82fc7c1c8271c7edc5 does a little better and drops onevent style attributes before outputting the element.

We can't solve every problem an extension will encounter in base, but I think this extra check is definitely a worthwhile since we *can* do it 😃 !

aidantwoods:
Btw, incase anyone is a little dubious about just checking for the `on` prefix in an attribute name solving this. Per w3 HTML5 spec, that should cover all event handlers.

https://www.w3.org/TR/html5/webappapis.html#event-handler-attributes
> An event handler has a name, which always starts with "`on`" and is followed by the name of the event for which it is intended.

tillkruss:
Code-blocks like will remain untouched, right?

m1guelpf:
This should be merged ASAP...

Parsedown.php
```diff
@@ -75,6 +75,25 @@ function setUrlsLinked($urlsLinked)
...
+    function setSafeLinksEnabled($safeLinksEnabled)
+    {
+        $this->safeLinksEnabled = $safeLinksEnabled;
+
+        return $this;
+    }
+
+    protected $safeLinksEnabled = true;
+
+    protected $safeLinksWhitelist = array(
```
erusev: Does this have to be an instance member? Seems like it can be moved to `filterUnsafeUrlInAttribute`

aidantwoods: In theory it would let an extension define additional protocols without having to worry about implementation of how to whitelist them

aidantwoods: I.e. you could fix a bug in the filter implementation that wouldn't have to be reflected in the extension if they're just adding data.

aidantwoods:
@tillkruss 
> Code-blocks like will remain untouched, right?

On a high level yes, they would go untouched. As-in the same markdown will be converted to the same HTML.
Though, all encoding is now done by the element method (and no longer by the particular block/inline methods).

For this reason it would *technically* be non backwards compatible on the `protected` API level (e.g. an extension has implemented their own code block methods which encode the output, and as a result of the base class changing, output would now be double encoded).

Something @erusev might want to consider when deciding the version level to increment.

Parsedown.php
```diff
@@ -91,7 +91,9 @@ function setSafeLinksEnabled($safeLinksEnabled)
         'ftp://',
         'ftps://',
         'mailto:',
-        'data:image/png;',
+        'data:image/png;base64,',
+        'data:image/gif;base64,',
+        'data:image/jpg;base64,',
```
tillkruss: It's actually `image/jpeg` ;-)

aidantwoods: Cheers ;-)

Is jpg allowed as a variant? (i.e. should I add both?)

tillkruss: No, it's invalid.

aidantwoods: Cool, ty! 😃

aidantwoods:
@erusev I've included the closes keyword at the bottom of the initial PR description on a few issues and PRs that should be made obsolete on merge. Might want to double check they close automatically though, since they were added by editing the message (not sure of the mechanics of doing that).

Parsedown.php
```diff
@@ -75,6 +75,27 @@ function setUrlsLinked($urlsLinked)
...
     protected $urlsLinked = true;
 
+    function setSafeLinksEnabled($safeLinksEnabled)
+    {
+        $this->safeLinksEnabled = $safeLinksEnabled;
+
+        return $this;
+    }
+
+    protected $safeLinksEnabled = true;
```
PhrozenByte: IMO this shouldn't be enabled by default, it breaks BC. Enabling this without `setMarkupEscaped(true)` makes no sense anyway - it prevents absolutely nothing without escaped markup. Never confuse people by letting them do pointless things.

How about something like this (disallow enabling `safeLinksEnabled` without `markupEscaped` and always enable/disable `safeLinksEnabled` and `markupEscaped` at the same time; however, disabling `safeLinksEnabled` when `markupEscaped` is enabled is still possible - escaping markup doesn't necessarily imply a "safe mode", but in contrast, a "safe mode" isn't possible without escaping markup)?

```php
    function setMarkupEscaped($markupEscaped)
    {
        $this->markupEscaped = (bool) $markupEscaped;
        $this->safeLinksEnabled = (bool) $markupEscaped;
        return $this;
   }

    function setSafeLinksEnabled($safeLinksEnabled)
    {
        // or throw a exception instead
        $this->safeLinksEnabled = $this->markupEscaped && $safeLinksEnabled;
        return $this;
    }
```

Parsedown.php
```diff
@@ -1485,10 +1503,80 @@ function parse($text)
...
+
+    protected function filterUnsafeUrlInAttribute(array $Element, $attribute)
+    {
+        if ($this->safeLinksEnabled)
+        {
+            $safe = false;
+
+            foreach ($this->safeLinksWhitelist as $scheme)
+            {
+                if (stripos($Element['attributes'][$attribute], $scheme) === 0)
```
PhrozenByte: I didn't test this, but I think it treats relative links like `[some link](relative/path/to/page)` as unsafe.

Parsedown.php
```diff
@@ -1485,10 +1503,80 @@ function parse($text)
...
+        {
+            foreach ($Element['attributes'] as $att => $val)
+            {
+                # clear out nulls
+                if ($val === null)
+                {
+                    unset($Element['attributes'][$att]);
+                }
+                # filter out badly parsed attribute
+                elseif (strpbrk($att, $badAttributeChars) !== false)
```
PhrozenByte: What's the reason for this? Letting badly parsed attributes produce gibberish is IMO better than silently ignoring them, its more explicit and gibberish actually helps people to understand that something went wrong. :wink:

Parsedown.php
```diff
@@ -1485,10 +1503,80 @@ function parse($text)
...
+        }
+
+        if ( ! empty($Element['attributes']))
+        {
+            foreach ($Element['attributes'] as $att => $val)
+            {
+                # clear out nulls
+                if ($val === null)
+                {
+                    unset($Element['attributes'][$att]);
```
PhrozenByte: What's the reason for this?

Parsedown.php
PhrozenByte: Furthermore: Don't use `stripos()` here, it forces PHP to walk through the whole string even when the first character already didn't match. Use `substr()` instead.

Parsedown.php
```diff
@@ -1485,10 +1503,80 @@ function parse($text)
...
+                }
+                # filter out badly parsed attribute
+                elseif (strpbrk($att, $badAttributeChars) !== false)
+                {
+                    unset($Element['attributes'][$att]);
+                }
+            }
+
+            $onEventAttributeKeys = preg_grep('/^on/i', array_keys($Element['attributes']));
```
PhrozenByte: Expensive and unnecessary approach. You're iterating through all attributes anyway, why don't you simply test for attributes starting with "on" using `substr()`?

Parsedown.php
aidantwoods: Yes, so some of this is spoken to by decisions made in the original PR.

I think the default was @erusev's suggestion as far as I can tell though, see https://github.com/erusev/parsedown/pull/276#issuecomment-71331724

If it was up to me I'd make `markupEscaped` default to on too 😉

I'm happy to go with the flow here though, so I'll await some more feedback

Parsedown.php
aidantwoods: Was needed when I used `array_flip` in place of where `array_keys` is now, can remove that now though :)

Parsedown.php
aidantwoods: So I'd rather throw an exception here TBH, but there isn't really precedent so for in Parsedown for throwing them (so I'll defer that decision to @erusev)

In my experience coming across this stuff in the wild, there's a pretty high chance that people will not test their applications with bad data (especially if they don't know it could even be valid).

Just for reference, the only chars in that list that would be "valid"-(ish) in the attributes are `"'\0`, the rest are whitespace (so impossible).

All the whitespace characters are invalid in an HTML tag's attribute name, and would certainly create two attributes (so very likely there is an attack).

By filtering out attributes containing these characters we only need check for a prefix at the start of the string since we guarantee that there can not be multiple attributes (as interpreted by the browser) contained in the single attribute we are considering.

Parsedown.php
aidantwoods: Yup you're right about that, needs to be case insensitive though. Is `strtolower` + `substr` going to be faster than `stripos`?

Parsedown.php
aidantwoods: Yup, it'll treat that as unsafe. Problem with allowing `relative/path/to/page` will be that you'll additionally have to check that no colons are present (also note that `&colon;` is equivalent to a colon in a HTML attribute, and contains no colons).

Also see https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet#Embedded_newline_to_break_up_XSS for some examples of getting XSS in a HTML attribute via some obscenely permissive browser parsing :(

e.g. `javas\0cript:` and `jav\tascript:` are valid :(

Parsedown.php
aidantwoods: Btw this isn't technically exhaustive if it happens to be the first attribute, see https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet#Non-alpha-non-digit_XSS

aidantwoods: *sigh* I guess we'll also have to ensure that the attribute name starts with an alphanumeric.

Just tested and chrome parses `<a /onmouseover="alert(1)">xss</a>` as  `<a onmouseover="alert(1)">xss</a>`

aidantwoods: So https://github.com/erusev/parsedown/pull/495/commits/4bae1c9834382d3c7aa900a7cbf77771b3864c56 implements this via a whitelist (which is safer), though @erusev should decide whether this is worth the performance hit given Parsedown base doesn't allow user generated attributes vs. allowing extensions to basically no checking on what attributes they allow, but Parsedown handling them safely.

Parsedown.php
aidantwoods: As an alternate method for what's done here, one idea I've had is selectively partially url encoding attributes that can't be guaranteed safe by the whitelist.

I've written an implementation [here](https://github.com/Parsemd/Parsemd/blob/master/src/DisplayAsHtml.php#L111-L140). One thing I wasn't sure about was how permissive to be with which characters to allow though. I think holding back on encoding`/#?&=` (so URLs will still function) (but definitely encoding `;`, `:` and all other non-alphanumerics) should work. I'll have to have a think about it a bit more though as to whether you could bypass that and specify a protocol.

PhrozenByte: Nevertheless are relative paths an important use case. It might be a good idea to look into the specs to find out how URI schemes must look like to be valid. Just guessing here, but I think the `:` as scheme "indicator" works only until a `/` is encountered. As an alternative: If safe url parsing isn't enabled by default, it might be sufficient to document this behavior and to tell people to use `./` for relative paths.

aidantwoods: > Nevertheless are relative paths an important use case.

Indeed. Though blocking them comes from the original PR, and there's a pretty good reason for it (non trivial case to distinguish from a protocol definition). 

> It might be a good idea to look into the specs to find out how URI schemes must look like to be valid. Just guessing here, but I think the : as scheme "indicator" works only until a / is encountered. 

Simply looking for `:` isn't sufficient (since `&colon;` is interpreted as a colon in the attribute value). Neither is consulting the specs on a how a valid URL should look :( if you take a look at the page I linked to, you'll see many things that aren't valid URLs that browsers have been known to accept (I presume the browser's intent is to understand URLs less strictly by applying corrections). 

I can whitelist `./` too, but what do you think of the selective partial URL encoding approach I mentioned (as opposed to dropping the attribute entirely)?

aidantwoods: @PhrozenByte https://github.com/erusev/parsedown/pull/495/commits/054ba3c48729036fd8ab1166fb0b1d0b4dd1e465 should address relative URLs while keeping safeUrls on by default.

tillkruss:
Great work man!

Parsedown.php
```diff
@@ -1485,10 +1502,80 @@ function parse($text)
...
+                {
+                    $safe = true;
+
+                    break;
+                }
+            }
+
+            if ( ! $safe)
+            {
+                $Element['attributes'][$attribute] = preg_replace_callback(
```
erusev: Can we use an alternative of `preg_replace_callback` that doesn't require passing a function definition as a parameter? Thanks

aidantwoods:
https://github.com/erusev/parsedown/pull/495/commits/054ba3c48729036fd8ab1166fb0b1d0b4dd1e465 is borrowed from one of [my projects](https://github.com/Parsemd/Parsemd/blob/060eda3cf7308bba02ed8a45c4873b412a6b8e96/src/DisplayAsHtml.php#L111-L140), which isn't something that's got a release version yet – so we should add some tests to verify that this code does indeed have non breaking behaviour.

Parsedown.php
aidantwoods: The current way it's set out is to have a whitelist of characters we don't want to encode (and encode everything else).

I could achieve the same result with `strpbrk` and encoding done on ranges of characters, not sure that would be faster though. Unless you have another idea?

aidantwoods: Ah sorry, didn't see your edit.

Could do it with preg_match_all with match positions returned and applying encoding to those ranges, again, not sure it'll be faster?

P.S. the regex will match as many chars as it can in one go before making the function call at present

aidantwoods: @erusev this would be an alternative using `preg_match_all`

```php
if (
    preg_match_all(
        '/[^\/#?&=%]++/',
        $Element['attributes'][$attribute],
        $matches,
        PREG_OFFSET_CAPTURE
    )
) {
    $offset = 0;

    foreach ($matches[0] as $match)
    {
        $len = strlen($match[0]);
        $encoded = urlencode($match[0]);

        $Element['attributes'][$attribute] = substr_replace(
            $Element['attributes'][$attribute],
            $encoded,
            $offset + $match[1],
            strlen($match[0])
        );

        $offset += strlen($encoded) - $len;
    }
}
```

More likely to introduce bugs with this though IMO, and I'm not sure it'll save that much on performance just looking at it.

Parsedown.php
PhrozenByte: > If it was up to me I'd make markupEscaped default to on too 😉

My :-1: for this, BC is IMO very important 😝

Parsedown.php
PhrozenByte: 👍

Parsedown.php
PhrozenByte: Definitly! `preg_match` (as in 4bae1c9) is very slow either. Starting the PCRE engine is expensive!

Parsedown.php
aidantwoods: Hehe, I mean as an initial decision (of course too late now! ;p)

Parsedown.php
PhrozenByte: I'm still not happy with this solution... Way too many false positives here...

How about blacklisting malicious protocols? Sure, browsers try to be as less strict as possible, but by removing any non-alphanumric chars and comparing the result against "javascript" you should catch all attacks with basically no false positives. Have you checked how other frameworks (not necessarily Markdown parsers, basically any comment section deals with this problem) handle this?

Parsedown.php
aidantwoods: So are we proposing?

```php
# dump onevent attribute
elseif (strlen($att) > 1 and strtolower(substr($att, 0, 2)) === 'on')
{
    unset($Element['attributes'][$att]);
}
```

instead of

```php
# dump onevent attribute
elseif (stripos($att, 'on') === 0)
{
    unset($Element['attributes'][$att]);
}
```
?

Parsedown.php
PhrozenByte: I'm pretty sure `preg_replace_callback()` is the easiest and fastest solution here :wink:

Parsedown.php
aidantwoods: `[xss](javas&#x0A;cript:alert(1))` will fail on stripping non-alpha numerics but is still xss, feel free to check them ;-)

aidantwoods: I keep posting this link, but I'm not convinced you've seen the examples 😉

aidantwoods: I think if you HTML decode, then strip out all non alphanumerics, then check for `javascript:` then HTML encode again, then that might do it. ;p

aidantwoods: > Encoding is no real solution as it breaks relative paths

Btw encoding will **not** break relative paths, please see the commit message in https://github.com/erusev/parsedown/pull/495/commits/054ba3c48729036fd8ab1166fb0b1d0b4dd1e465

I'll have to have a think about that HTML decode, then running the check and whether that would hold up as a solution.

aidantwoods: > How about blacklisting malicious protocols?

Blacklists are hard to get right

PhrozenByte: I've seen the examples :wink: But there must be a better approach. I don't think that other software deals with that problem in this way. However, I definitely was wrong by assuming that encoding breaks relative paths. It *depends*. You obviously must not encode `/`, but everything else seems to be fine when encoded (i.e. your solution works). Nevertheless, again, I don't think that re-inventing the wheel makes sense here. How does Symfony encode URLs? Or Zend? Or end-user software that accepts arbitrary URLs like phpBB, MediaWiki, WordPress? The list is basically endless.

aidantwoods: I'm happy to switch to something you can point to that works 😉

Generally speaking, whitelists are safer than blacklists (i.e. try and figure out what to allow, rather than what to block). So even if the HTML decode method I suggested above seems to work, I'd personally still opt to use a whitelist over it. (since it's just a fancy blacklist).

aidantwoods: Looks like Symfony urlencodes 😉 

screen shot 2017-05-04 at 00 10 28

PhrozenByte: \**Striking through what I've written before as you're checking Symfony right now\**

I don't think that we're talking about the same thing, especially regarding whitelists and blacklists. You're talking about sanitizing and validating input (something that is crucial!), not whitelisting. I'm in contrast talking about whitelisting particular protocols like `http://` and `ftp://`, as this is, by definition, something user's will stumble upon (because there's nothing bad about `irc://`, `steam://` and so on - the list is endless, that's problem of trying to whitelist them)

However, it doesn't matter. I think that you'll get enough input by looking at how Symfony solves this problem ([as said](https://github.com/erusev/parsedown/pull/495#discussion_r114670410), encoding is probably the right solution in general, but there are still some open questions) - so I'll simply wait for your next update :smiley:

aidantwoods: > I don't think that re-inventing the wheel makes sense here. How does Symfony encode URLs? Or Zend? Or end-user software that accepts arbitrary URLs like phpBB, MediaWiki, WordPress? The list is basically endless.

A lot of software that accepts input from (untrusted) users probably won't have any reason to support relative URLs

e.g. phpBB doesn't support them 🤷‍♂️ 
screen shot 2017-05-04 at 00 33 56

---


> I'm in contrast talking about whitelisting particular protocols like http:// and ftp://, as this is, by definition, something user's will stumble upon (because there's nothing bad about irc://, steam:// and so on - the list is endless, that's problem of trying to whitelist them)

Unfortunately there's probably an endless list of ways to specify what might be useful to henceforth call a 'dirty' protocol (that isn't technically valid, but browsers will accept).
And telling the difference between a dirty protocol and a relative URL is non trivial.

I haven't worked with Symphony's internals before, so tracking down exactly how links are built in full would probably take a while for me to do.
Safe to say they do not do it by searching for the string `javascript` though 😉 

screen shot 2017-05-04 at 00 51 58

aidantwoods: IMO we should make an effort to expand the list of allowed protocols as much as possible and ones that we don't know about can be added as bugfixes as we go.

And forgetting to support something like `steam://` IMO is more favourable to accidently letting someone's website become vulnerable to XSS because we didn't get the filter quite right, especially when a user could easily just leave the URL as plain text as a work around.

aidantwoods: Also, I'd just like to point out a great example of why blacklists are hard to get right

> > Blacklists are hard to get right
> That's not true. javascript: is the only protocol you must detect and remove to prevent XSS attacks. Everything else is a false positive.

`data:text/html,<script>alert(1)</script>` is xss

😉

aidantwoods: [boo](http://google.com)


GitHub uses a whitelist

screen shot 2017-05-04 at 01 20 18
screen shot 2017-05-04 at 01 20 22

vs.

screen shot 2017-05-04 at 01 21 09
screen shot 2017-05-04 at 01 21 14

PhrozenByte: I don't think that I've something to add at the moment besides repeating myself, so I'm just waiting for your updated solution :smiley: Notwithstanding our discussion, you're doing a great job!

aidantwoods: So, standing back and looking at the full picture here...

We can actually dispose of worrying about the `&colon;` (`&#58;` etc...) cases entirely because everything gets HTML encoded at the end anyway 😜 (`&`is not allowed).
So all HTML entities will become double encoded prior to being output, so this would be the only way of specifying a protocol.

Furthermore, just to make a correction to the commit message in https://github.com/erusev/parsedown/pull/495/commits/054ba3c48729036fd8ab1166fb0b1d0b4dd1e465 .
It (upon further testing) is in-fact possible to generate a colon in an HTML attribute using just the characters `&` and `#`. Specifically, while Chrome does not permit `&colon` (without `;`), it does permit `&#58` 😞.

As said, we're still safe because of the HTML encoding done on all elements though.

aidantwoods: So I remembered that many years ago (when I used Windows 😉 and knew some VBScript), that IE supported the URI.

Check out [here](https://books.google.co.uk/books?id=Znxa3zrJWJsC&pg=PA239&lpg=PA239&dq=internet+explorer+vbscript+uri&source=bl&ots=0I3ow9kayV&sig=CTUbiU2Q9VZtIUkwTT47M5dW6Ec&hl=en&sa=X&ved=0ahUKEwjWw7uUxdnTAhUJ2RoKHY9BCRAQ6AEIQjAF#v=onepage&q=internet%20explorer%20vbscript%20uri&f=false), and [here](https://gist.github.com/rvrsh3ll/09a8b933291f9f98e8ec#file-xxsfilterbypass-lst-L48-L49).

I just don't think it's possible to allow custom protocols *and* be free from XSS via navigation.

aidantwoods: Even if I could enumerate all the protocols that allowed scripting, what if a new one comes into existence?

I started trying to write this to give us a bigger whitelist

```bash
curl -s https://www.iana.org/assignments/uri-schemes/uri-schemes.txt | egrep 'Permanent|Provisional|Historical' | cut -d ' ' -f 1 | sed "s/^/'/" | sed "s/$/:',/"
```

But `data:` is listed in there (which makes sense I suppose).

So it then changed to this (addition of `egrep -v '^data|^javascript|^$'`)

```bash
curl -s https://www.iana.org/assignments/uri-schemes/uri-schemes.txt | egrep 'Permanent|Provisional|Historical' | cut -d ' ' -f 1 | egrep -v '^data|^javascript|^$'  | sed "s/^/'/" | sed "s/$/:',/"
```

But really, I'm just moving the blacklist somewhere else by doing this. There's no way I can honestly go through all those protocols and decide whether or not they would permit scripting in certain situations.

I think the only thing we can sensibly do is start with some common protocols on the whitelist, and then add new ones after a bug report is issued, and someone evaluates whether adding that protocol would permit scripting.

Parsedown.php
aidantwoods: How's this https://github.com/erusev/parsedown/pull/495/commits/edffa208854610b3a6d5b9ff99fed19f7c0cbc7c ?

Parsedown.php
aidantwoods: @erusev as far as I can see it should be safe to just url encode the colon character and nothing else in order to break custom protocols (i.e. a `str_replace` instead of that `preg_replace_callback`).

Only as far as I can see though – the `preg_replace_callback` enforces a whitelist on characters we allow ***not*** to be URL encoded, whereas only encoding that one character would let anything through.

I'd advise that we keep this character whitelist in place, and URL encode everything we haven't thought of. That said, it would be a slight performance trade off (in the special case of relative URLs), so I'll go with the flow here (seeing as I can no longer see any concrete way to bypass just encoding the single character).

aidantwoods: https://github.com/erusev/parsedown/pull/495/commits/dc30cb441c0834357b7f145444f6e53cbe67154e expands the whitelist to allow some more common-ish protocols.

aidantwoods:
Hmm.. looks like travis is having some internal issues with hhvm here. Remaining tests are passing

screen shot 2017-05-06 at 17 24 09

Parsedown.php
```diff
@@ -1485,10 +1510,94 @@ function parse($text)
...
+    {
+        if ($this->safeLinksEnabled)
+        {
+            $safe = false;
+
+            foreach ($this->safeLinksWhitelist as $scheme)
+            {
+                if (self::striAtStart($Element['attributes'][$attribute], $scheme))
+                {
+                    $safe = true;
```
PhrozenByte: Simply return immediately (`return $Element;`), there's nothing more to do here.

m1guelpf:
@aidantwoods Is this being merged soon?

Parsedown.php
aidantwoods: Yup, good catch!
226f636360d29060f9d6d15ec47ce5de575a226c

Parsedown.php
PhrozenByte: I'm still not entirely convinced that this is best solution in every detail, but you've convinced me that the approach is right. There might still be some open detail questions, I unfortunately don't have time to look through everything in detail at the moment.

aidantwoods:
@m1guelpf hopefully, there are a couple things I'm awaiting feedback on (cc: @erusev)

---

@erusev I realise the comment threads on this one are a little hard to track down 😉 I've cc'd you into a couple of them where I'm waiting for feedback, but just so there's a bit of an easier navigation point:

Should we filter out event handler attributes? A performance trade off that Parsedown doesn't strictly need now, but offers safety for future if Parsedown/Parsedown extra ever permits user generated attributes, and immediate benefit to extensions that permit this now (can point to at least one example).
https://github.com/erusev/parsedown/pull/495#discussion_r114450311

Should we URL encode just the colon `:` or maintain a whitelist of characters we do not wish to URL encode? (This with the aim of preventing a protocol/scheme definition). Regardless of the approach taken here, HTML attributes will always be properly escaped because we HTML encode elsewhere (preventing use of HTML entities, and breaking out of the quotes).
I can't see a concrete reason that we'd need to use the whitelist at present, it would be a precaution (but with a non-insignificant performance trade off in the case that a defined URL does not match a whitelisted protocol/scheme).
https://github.com/erusev/parsedown/pull/495#discussion_r115087339

Parsedown.php
aidantwoods: What if we used your approach https://github.com/erusev/parsedown/pull/495/files/aee3963e6b97186b1e5526c118bf5d2d872cd8ee#r114438464 to make `safeLinks` be off by default, and strongly coupled `safeLinks` and `setMarkupEscaped` to be either both on, or both off?

That way custom protocols are supported for authors (who trust their own content), but will be off for user generated content (that they do not trust).

Perhaps we could ease the complication of having to enable two options in sync by introducing a new option that does both behind the scenes: `preventXss` or `setContentUntrusted` – that way we'd be giving a more explicit representation to the user about what this option actually does, and would afford us the assumption that while in this mode we should treat content as untrusted.

PhrozenByte: The one has nothing to do with the other, never ever introduce BC breaking changes without incrementing the first digit of the version (i.e. when this feature is enabled by default, it **must** be released as Parsedown 2.0). You'll break **all** dependent applications otherwise - and Parsedown is no end user project, it's a library, thus you'll break hundreds of applications with thousands of users. See http://semver.org. The only exception is behavior that never was expected (like some weird edge cases when parsing markup), neither officially, nor how it was practically used by others.

aidantwoods: > never ever introduce BC breaking changes without incrementing the first digit of the version (i.e. when this feature is enabled by default, it must be released as Parsedown 2.0)

I'd tend to agree with that. In-fact even if we change no public default behaviour, IMO it would be valuable to consider how much of a version jump changes to the protected API would warrant (see: https://github.com/erusev/parsedown/pull/495#issuecomment-298717654)

What do you think about introducing a new option though – in theory if we introduced something like `preventXss` or `setContentUntrusted` then there would be no BC breaking changes to the public API (we then wouldn't need to strongly couple `safeLinks` and `setMarkupEscaped` into a both or none scenario – perhaps we wouldn't even need to expose `safeLinks` publically, but just the new option).

PhrozenByte: If it is disabled by default - sure, why not.

However, as I've said before (https://github.com/erusev/parsedown/pull/495#discussion_r114438464), please keep in mind that people might want to escape markup without having security/user-generated content in mind. Thus there might still be users who do want to escape markup, but don't want to have `safeLinks` enabled (i.e. what Parsedown currently does when `markupEscaped` is enabled). Just the other way round (`safeLinks` enabled without `markupEscaped`) makes no sense.

Nevertheless, this is no crucial question for me. Important is that this isn't enabled by default, as this would break BC in an absolute catastrophic way.

aidantwoods:
I've made a couple changes to this branch, each applying to a possible response to the above questions. I'll write a little about them here:

---

https://github.com/erusev/parsedown/pull/495/commits/b1e5aebaf6c8162cda482e96a1370155fd2162b5 follows on from [#495 (comment)](https://github.com/erusev/parsedown/pull/495#discussion_r115144576).

This acts to **disable** safe links by default. As pointed out by @PhrozenByte it doesn't really make sense to have it enabled by default, given that `setMarkupEscaped` is disabled by default (you need both enabled to benefit).
However, this is good because we'll now have a setting that lets a user of Parsedown express whether or not they would like the output content to be free of XSS. (therefore anything that is XSS is a bug, and can be fixed in a backwards compatible manner without needing to introduce lots of confusing options for each anti-xss feature we create).

This also means that this PR should now be **non** backwards compatible breaking on the public API level (though on the protected level there is a BC break because encoding is handled later than previously in order to have it apply to all elements, see [#495 (comment)](https://github.com/erusev/parsedown/pull/495#issuecomment-298717654))

---

https://github.com/erusev/parsedown/pull/495/commits/e39085ca38ffd56e6b92ac5c4371cc7441912cdc follows on from [#495 (comment)](https://github.com/erusev/parsedown/pull/495#discussion_r115087339) .

Since I can't think of a good reason to encode more than the colon when applying sanitisation procedures (note that HTML entities will be un-insertable because we HTML encode in a different step), I've remove the `preg_replace_callback` that whitelists allowed characters, and have opted to only encode the single `:` character, since https://tools.ietf.org/html/rfc3986#section-3 indicates that it is required syntactically, and so browsers *should* not recognise a protocol/scheme if the entire URI contains no colons.

This behaviour is governed by `safeMode` as above, meaning all links are possible by default (incl. `javascript:`), so this mode should be enabled to take advantage of the behaviour previously called `safeLinks`.

---

With regard to [#495 (comment)](https://github.com/erusev/parsedown/pull/495#discussion_r114450311) and filtering out on-event attributes, this will occur iff `safeMode` is enabled.

hkdobrev:
@aidantwoods I wouldn't have the time to send a PR to your branch with more test cases, but you might find more XSS examples on the OWASP website:
- https://www.owasp.org/index.php/Testing_for_Reflected_Cross_site_scripting_(OTG-INPVAL-001)
- https://www.owasp.org/index.php/Testing_for_Stored_Cross_site_scripting_(OTG-INPVAL-002)
- https://www.owasp.org/index.php/Testing_for_DOM-based_Cross_site_scripting_(OTG-CLIENT-001)

Cheers! 🍻

aidantwoods:
@hkdobrev cheers for those, that first link has been my reference point more than a couple of times for what to defend against

As far as I can tell the tests I've added should encompass the spirit of most of the applicable ones from filter evasion cheat sheet.

Also from OWASP, Zap is unable to find XSS with `setSafeMode(true)` on.

Using the following test file:
```php
<?php

require('Parsedown.php');

$p = new Parsedown;
$p->setSafeMode(true);

echo $p=>text($_GET['text'])."\n";
echo $p=>text('[link]('.$_GET['text'].')')."\n";
echo $p=>text('['.$_GET['attribute'].']('.$_GET['text'].')')."\n";
```

By default, Zap will find all three versions of XSS that I'm hinting it can find (by writing the markdown for it around the paramaters):
screen shot 2017-05-10 at 19 40 19
screen shot 2017-05-10 at 19 40 24
screen shot 2017-05-10 at 19 40 31


But with `setSafeMode(true)`, the same test file gets this (down at the bottom XSS is absent from alerts):
screen shot 2017-05-10 at 19 41 17

m1guelpf:
@aidantwoods Then, this is ready?

aidantwoods:
Thanks guys for all the feedback!

---

@m1guelpf I'm happy with it, so good for merge from me!

---

@erusev once #499 is merged into master Travis should give this the okay (hhvm won't build on current master)

aidantwoods:
Now that #499 has been merged, the checks should pass if the last travis build is restarted.

m1guelpf:
@aidantwoods Any updates?

aidantwoods:
> @aidantwoods Any updates?

/fwd: @erusev

erusev:
I need a little more time.

aidantwoods:
@erusev anything that's preventing merge that I can do to help?

aidantwoods:
Hello everyone!

I'm not sure why the merge here is taking so long, or if it'll ever get merged – let alone into a release :(

So that everyone can start benefiting from this right away, I've created a class extension to Parsedown and made it available via composer.

Just 

```bash
composer require aidantwoods/secureparsedown
```

to grab it.

Extension repo here: https://github.com/aidantwoods/SecureParsedown

To use it just,

```php
<?php

use Aidantwoods\SecureParsedown\SecureParsedown;

$Parsedown = new SecureParsedown;
$Parsedown->setSafeMode(true);
```

And then how you'd normally use Parsedown from then on out, enjoy!

Daniel-KM:
@erusev 
Parsedown and parsedown-extra are now implemented by default in some common cms and tools. This is why there are a lot of forks and stars. So this may become difficult to maintain for one maintainer only. So the repository may be opened (write rights) a little more to people that are very interested by this parser.

m1guelpf:
@Daniel-KM I think that deserves another issue...

Daniel-KM:
Yes, I'm going to open a new one. I wrote here, because security should be the first thing to think.

dregad:
Ping @erusev

h4ckninja:
Good stuff. A quick test case seems to pass:

~~~php
<?php

require('./../vendor/autoload.php');


use Aidantwoods\SecureParsedown\SecureParsedown;


// Initialize.
$parsedown = new SecureParsedown();
$parsedown->setSafeMode(true);
$parsedown->setBreaksEnabled(true);


// XSS through href.
$url = '[Click me - I\'m a JavaScript alert.](javascript:alert(\'Hi%20there.\'))';
$xss_url = $parsedown->parse($url);
echo $xss_url;


// XSS through attribute injection on an href.
$attribute = '[Click me - I\'m breaking out to an attribute.](http://www.example.com/"onclick="alert(\'Hi.\');)';
$xss_attribute = $parsedown->parse($attribute);
echo $xss_attribute;

// Code block
$code = '~~~php
<?php

phpinfo();
';
$parse_code = $parsedown->parse($code);
echo $parse_code;


// HTML
$html = '<strong>hi</strong>';
$html_code = $parsedown->parse($html);
echo $html_code;
~~~

tillkruss:
Is this package dead?

erusev:
@tillkruss 

No, not at all.

embluk:
@erusev Come on man, you have something good going on here, a great parsedown in PHP. This XSS vulnerability is pretty serious for everyone using your project...

erusev:
@embluk 

I understand that, but please consider that:
1. It's a parser and not a sanitizer, which doesn't mean that it shouldn't sanitize, I personally think it should sanitize, because it's more efficient for sanitization to happen inside of the parser rather than outside, but still, the main purpose of the library is to sanitize
2. I can't merge this and announce that the library is safe to use without understanding the problem and the solution super well, it would be irresponsible.

Perhaps, until the issue is resolved, I should add a note to the readme that says that the library is just a parser and that, when necessary, people should use a sanitizer along with it.

embluk:
@erusev 
I guess you're right in it being a parser, but yeah adding a note might help new people realise what they need to do in order to use it safely in the wild. Something like HTML Purifier is what I've turned to and just set what HTML is allowed based on what this parser outputs.

Yeah I understand and I'm glad you wish to understand the problem before the solution first, god speed.

erusev:
I'll try to make some time in the next couple of week to deal with the issue.

I'm also working on 2.0 that will introduce a simpler, more modular structure of the code, hopefully, I'll be able to release it by the end of the year.

erusev:
Just added a security section to the readme to include the note I mentioned above.

embluk:
@erusev Very good, thanks for your quick responses. I'm sure it will help others and guide them in the right direction to use this parser safely. Looking forward to the future updates as well ;D

aidantwoods:
@erusev 

> 2. I can't merge this and announce that the library is safe to use without understanding the problem and the solution super well, it would be irresponsible.

Often a good way to understand something is to ask questions, so please don't hesitate to ask me or others questions if it'll help.

Something else to note though – in the world of security no one who doesn't want to eventually be proved wrong will ever claim that a solution is "100% safe". This something that is perhaps obvious as we see more and more security incidents making the news these days. You're never going to be able to make the claim that the library is safe in a binary safe/not safe sense. (In-fact, not even pen-testing consultancies who *literally do security testing for a living* will claim that the software they have tested is guaranteed to have no issues, even when they don't find any).

That might seem dire, but not all is lost. This is why software is often pen-tested to make sure that we have at least learnt and applied lessons from past mistakes. In the case that a new or unnoticed exploit becomes apparent, the best tool that you can have at your disposal is the ability to push out a fix and get users updating as quickly as possible.

In short, go ahead and try to understand the PR as best you can – you're right in that you should be able to be confident that it works. Just be careful not to confuse confidence with certainty – you can always fix new issues, but if you're waiting around for it to be provably secure before release then no one will ever get an update.

h4ckninja:
> Often a good way to understand something is to ask questions, so please don't hesitate to ask me or others questions if it'll help.

Hi. Real-life pentester here. I've gained access to fortune 500's (with permission of course), have broken software, and have contributed to the open source community through my expertise in security. I am more than happy to answer questions as well.

"It's just a parser" is, to be honest, a dangerous mindset. Let's take another parser that had a significant code flaw that I found:

https://github.com/dedalozzo/pygmentize/issues/1

This flaw resulted in Remote Code Execution on the victim server because it was parsing user input to provide highlighting of source code.

Another example:

https://github.com/ryakad/pandoc-php/issues/15

I was able to get Remote Code Execution because the library parsing user input passed it in an unsafe manner. While the author took a long time to respond, the users of this popular library were left to be victims.

Another parser (someone else found this one):

https://github.com/Gregwar/RST/pull/34

This library was vulnerable to Local File Inclusion because the parser did not consider secure design principles.

My point here is not to pick on the amazing work you've done, @erusev, or anybody else's hard work. But that waiting on proven security, while leaving a library vulnerable, hurts users. I use this library for my projects because I like it. But leaving a parser to be "just a parser" is, honestly, dangerous. I've given a couple of examples above when secure design principles aren't followed.

I am more than happy to help in any way I can.

tillkruss:
@erusev: Thanks for the update!

I'm currently using `ircmaxell/stauros` to sanitize the generated HTML and that package includes a similar warning:

>THIS IS AN EXPERIMENTAL LIBRARY, USE AT YOUR OWN RISK
>— https://github.com/ircmaxell/Stauros

erusev:
@h4ckninja 

Thanks for the input! I understand that this is a serious issue and I'll work to resolve it.

aidantwoods:
Just to echo @h4ckninja's words here:
> waiting on proven security, while leaving a library vulnerable, hurts users. I use this library for my projects because I like it. But leaving a parser to be "just a parser" is, honestly, dangerous. I've given a couple of examples above when secure design principles aren't followed.

Genuinely every day that this goes unfixed in core, users apps remain vulnerable. Despite the (recently added) note in the README, the Wiki has read the following [since 2015](https://github.com/erusev/parsedown/wiki/Tutorial:-Get-Started/f9d881da38d8ef737ffc0c0a140b124ba3367904):
> ```php
> echo Parsedown::instance()
>    ->setMarkupEscaped(true) # escapes markup (HTML)
>    ->text("<div><strong>*Some text*</strong></div>");
> 
> # Output:
> # <p>&lt;div&gt;&lt;strong&gt;<em>Some text</em>&lt;/strong&gt;&lt;/div&gt;</p>
> ```

Giving a strong impression that the library is capable of escaping HTML. I mean, it's literally what it says – **the fact that it doesn't properly is an implementation detail *and a bug*** (there are other ways to achieve xss that this PR also solves that wouldn't be tackled by just escaping HTML, but that's not really the point).
There are just a hair over 16,000 code results for "parsedown setMarkupEscaped", and you don't have to look very far for repos that are relying on it for sanitising user input. I would encourage you to have a browse.

https://github.com/search?utf8=%E2%9C%93&q=parsedown+setMarkupEscaped&type=Code

The point here is to try and drive home the point that, unfortunately best intentions aside, this issue really does affect users.

aidantwoods:
@erusev 
Just to get the ball rolling on why some of the changes exist, consider the following:
https://github.com/erusev/parsedown/blob/9cf41f27abd6708ffaeb90d477f91dd792c86a97/Parsedown.php#L1252-L1257

Clearly you tried to do some HTML escaping here.

The first thing that stands out to me is the fact that you missed applying escaping to the title, an attribute listed just three lines above the escaping. Thus at present even with `setMarkupEscaped` enabled we can do this:
screen shot 2017-10-30 at 23 14 34

I'm not trying to nit-pick by pointing this out, rather this is to demonstrate how easy it is to miss something when trying to enumerate where every attribute that needs escaping is set.

Similarly, the `alt` description in images is vulnerable:
screen shot 2017-10-30 at 23 16 54

This type of XSS is fortunately easy to define, because it breaks the AST rules that Parsedown is representing internally. To address this, I moved the escaping into the `element` method

https://github.com/aidantwoods/parsedown/blob/67c3efbea0d33c4433c6b18ed163b45c01395867/Parsedown.php#L1441-L1455

This ensures that element attributes and content remain enclosed in their respective contexts.
I also opted to use a different method to do the escaping because the existing one was not complete (essentially it didn't encode enough characters). Also within this `element` method, two different contextual escaping modes are used (the element content permits the use of raw quotes, however the element attribute does not).

If that makes sense I can perhaps try and explain the remaining changes? It is a fairly small patch, so I'd hope that it wouldn't take too long to explain what remains.

sergeysviridenko:
It's very important changing. When it'll be implemented to master branch?

erusev:
@sergeysviridenko soon, most likely in December

sergeysviridenko:
@erusev I hope and thank you in advance.

levacic:
As Laravel has included this package as a dependency [for a while now](https://github.com/laravel/framework/pull/16768) (more than a year), and the current version was added to the [SensioLabs Security Advisories database](https://github.com/FriendsOfPHP/security-advisories/pull/257), any Laravel project started in that timeframe has started throwing errors when running the [CLI Security Checker tool](https://github.com/sensiolabs/security-checker) (e.g. in a CI context).

Laravel users don't really have much control over Laravel's dependencies, and for most users switching frameworks on existing projects is not a realistic option, so I'd urge @erusev to merge this as soon as possible, because that's the most convenient solution in this situation. If you don't have time to maintain this project, that's okay and understandable, but in that case please add more maintainers who can help out - @aidantwoods seems like an obvious choice based on the effort he invested already in this project (also see #553).

I'd just like to note that the fact that this has gone unmerged for such a long time, and with some quick calculations based on https://packagist.org/packages/laravel/framework/stats - it means that there's over 17.5 million Laravel installations out there (I have no idea how many actual live projects this implies though), which contain this known vulnerability.

Bonscho:
After utilizing parsedown for some years as a dependency of a corporate application framework we will drop it today. After leaving this pr unmaintained for such a long period it is absolutely reasonable to list the package as vulnerable. So also the roave/security-advisories did. For us as framework devs this means, that about 100 customer projects are undeployable due to the reluctance of a single person. 
@aidantwoods has our respect and compassion for the effort he invested.

tillkruss:
Replacing Parsedown as a dependency seems like a lot of effort. I'd rather see a forked repo that is maintained by the community.

Numline1:
+1 on this

aidantwoods:
@erusev You've made Parsedown available for free, and of course you're under no obligation to maintain the codebase on the behalf of those that choose to use it. However, there also is a large community which has chosen to depend on Parsedown. Unfortunately people cannot reasonably depend on software that receives no maintenance, and seems pretty clear to me that the way things are currently going, Parsedown therefore can't sustain the community that depends on it.

There seems to be an ongoing theme that, because you care about Parsedown, you want to carefully examine every line of code that goes into it. Unfortunately you also appear to be in a situation such that you cannot dedicate enough time into reviewing/writing code for Parsedown that would allow others to depend on it.

From my point of view there are two solutions to this that keep the community here:
1. You somehow find more time to support the community; or
2. You allow others to help out (this would mean permitting others to review, merge, and release code, as well as being able to manage issues – essentially some new admins on the repo).

Failing this I feel that Parsedown's community will look elsewhere, or worse: be left relying on stagnant code. Unfortunately I cannot personally recommend that anyone even use Parsedown currently: when security issues are left idle for so long it becomes dangerous to use software, especially if something worse than this were to come along – no one could be confident it would be fixed with any urgency. I would love to be able to recommend Parsedown if things change though.

I think the only other real solution failing the above two options is for some community members to proceed with new development on a fork.

@tillkruss 
> Replacing Parsedown as a dependency seems like a lot of effort. I'd rather see a forked repo that is maintained by the community.

Wherever new development ends up (including the possibility of it staying in this repo), #553 might be a good start for anyone willing to help.

donatj:
My pointless 2¢ but expecting your parser to also sanitize is asking too much. Sanitization is not a task to be taken lightly and should be left to an expert.

I'd say this is not a security issue at all, just an issue of misunderstanding the scope of the library.

aidantwoods:
> My pointless 2¢ but expecting your parser to also sanitize is asking too much. Sanitization is not a task to be taken lightly and should be left to an expert.

1. A parser may not be expected to sanitise, but if it has a sanitisation option it should work. Parsedown has an option that claims to escape HTML (this option is broken)

2. Parsedown introduces new XSS vectors that wouldn't otherwise be there. i.e.

```php
$parsedown =  new Parsedown;
$parsedown->setMarkupEscaped(true);
echo $parsedown->text(htmlentities($userinput));
```
is vulnerable, while the following generally isn't (if not echoed into an attribute, script tag etc...).

```php
echo htmlentities($userinput);
```

malas:
do the sanitation in your own (or your frameworks) code. 
this is just same old good best practice **separation of concerns**.

```php
$user_input = $request->get('markdown_from_a_form)) // this is the place where sanitation takes place.

$parsedown =  new Parsedown;
// $parsedown->setMarkupEscaped(true);  no need for that anymore at all

// return the response or for the sake of example, echo the markdown
echo $parsedown->text(htmlentities($user_input));
```

h4ckninja:
> My pointless 2¢ but expecting your parser to also sanitize is asking too much. Sanitization is not a task to be taken lightly and should be left to an expert.

I'm sorry, but I have to respectfully disagree. In this case, the library is the expert - it is consuming user input, parsing it, and outputting something different. It should handle that input securely. It is reasonable to expect the library to handle that securely. As @aidantwoods mentioned above, users of the library who are using the library may not fully understand the implications, and they won't be the experts of transformation from one input to another output - the library is.

aidantwoods:
@malas perhaps https://github.com/FriendsOfPHP/security-advisories/pull/257 and https://github.com/Roave/SecurityAdvisories/issues/44 may be useful references.

Separation of concerns is irrelevant because this is a bug in one of Parsedown's features. That's all there is to the discussion of it being a bug or not. If you disagree please create an issue/PR removing the `->setMarkupEscaped` feature.

malas:
@aidantwoods thanks.

i think removing `setMarkupEscaped` feature would be the best decision:

1. Less experienced people will not be fooled into thinking everything is secure.
2. Symfony, Laravel and other PHP frameworks have a built in user input sanitation and i do not think that there is a need for another one.

what do you think?

aidantwoods:
> i think removing `setMarkupEscaped` feature would be the best decision:

> 1. Less experienced people will not be fooled into thinking everything is secure.

This PR fixes all the XSS vulnerabilities that I am aware of and adds a new `setSafeMode` option to enable all security features (including sanitising of link destinations). I think the best place to fix this is in Parsedown, furthermore if you can achieve XSS with `setSafeMode(true)` enabled please let me know, I'd love to fix it. **This should be enough.**

> Symfony, Laravel and other PHP frameworks have a built in user input sanitation and i do not think that there is a need for another one.

Whitelisting of the output elements and attributes is probably the only other reasonable way to fix this, any sanitisation that occurs prior to passing input through Parsedown is flawed unless it is designed to sanitise against bugs in Parsedown, because Parsedown can introduce bad output if it is not designed to be safe.

> what do you think?

IMO it should be fixed in library.

h4ckninja:
>Symfony, Laravel and other PHP frameworks have a built in user input sanitation and i do not think that there is a need for another one.

Ah, I think this is the root of the confusion. This input that we are talking about isn't invalid input. It's the result of the transformation itself - that's why I argue that the parser should handle the validation of the input. The input only becomes dangerous because of the how the parser does its job. It's not a separation of concerns - it's the direct result of the library.

garygreen:
How does this fare against RSnake's famous cheatsheet of XSS attacks?

http://htmlpurifier.org/live/smoketests/xssAttacks.php

😄

eagostini:
For a while [parsedown/parsedown](https://github.com/parsedown/parsedown) has been kept as a copy of [erusev/parsedown](https://github.com/erusev/parsedown). Is it an option to carry the project on over there as an escape?

aidantwoods:
@garygreen I'd suspect pretty well, send me anything it fails on by [opening an issue here](https://github.com/aidanwoods/parsedown/issues). I have reviewed the patch against OWASP's list manually, and OWASP's Zap is unable to find XSS with safe mode enabled (a lot of this is discussed earlier in the thread).

Perhaps we can limit conversation here so that it stays on topic.

malas:
@h4ckninja my bad. i think i owe a beer for @aidantwoods :)

garygreen:
> I'd prefer not to have to argue about this being a bug here though.

Not sure why you thought I was raising an argument? I was simply asking how complex is this implemention and does it pass known XSS problems... I certainly don't want to stamp on your hard work here as you've put a lot of effort into the commit. I just wanted to highlight that XSS is a complex problem and people should be aware of all the exploits and that this PR might not solve every possible one (your more qualified to answer that question though) 😄

h4ckninja:
>my bad. i think i owe a beer for @aidantwoods :)

All good. :) This is why I think this discussion was useful - to make sure we all understand each other and what's going on.

aidantwoods:
@garygreen 
> Not sure why you thought I was raising an argument?

I apologise for the rushed messaged, I meant to tack the bit about arguing on at the end as as a more general point (but not directed at you) and realised it looked otherwise after posting. I sort of rephrased my comment in an edit afterward, which I hope it is a little friendlier. Again, sorry for the apparent hostility – my bad :)

aidantwoods:
@garygreen 

> I just wanted to highlight that XSS is a complex problem and people should be aware of all the exploits and that this PR might not solve every possible one

As far as I am aware this patch should solve all XSS issues in Parsedown (when safe mode is enabled), but I invite you (and others) to prove me wrong, and in that case I'll try my best fix whatever issues you find.

The idea behind safe mode is so that it is clear to the user when output is safe from untrusted user input (as opposed to introducing lots of confusing options like: remember to enable markup escaping, and link destination validation, plus all the new features we didn't realise we need to make the output safe that we're going to add in bug fixes etc..., or by just leaving the user to figure it out themselves and go wrong).

I prefer it be simple: If you can achieve XSS with safe mode enabled, it is a bug in safe mode. It should be fixed (and is also thus fixable without breaking backwards compatibility).

As a defence in depth measure, as a user you can also run something like HTMLPurifier on Parsedown's output, and deploy a CSP. But these should be backup measures, safe mode should do the heavy lifting.

dennishenry:
This issue is now causing a SensioLabs Composer Vuinerability Check to fail at our organization.

```
[ERROR] 1 packages have known vulnerabilities.                                 

erusev/parsedown (1.6.4)
------------------------

 * (no CVE ID): Cross-Site Scripting
   https://github.com/erusev/parsedown/pull/495
```

Is there any potential ETA on merging this and tagging a new release so this can be cleared? Thank you!

taylorotwell:
@erusev click merge. tag release. :clap:

taylorotwell:
@aidantwoods are these fixes merged on your fork? is your fork available on Packagist? Do you plan to maintain it?

dennishenry:
Thank you @erusev for merging this and your hard work on the utility as a whole!

erusev:
@taylorotwell

> @erusev click merge. tag release. 👏

Done.

I apologize for being so slow to react to this. I'll do my best to respond more quickly in future and I'll also consider adding more maintainers, @aidantwoods, are you interested?

stefandoorn:
Thanks @erusev!

dregad:
@erusev Thank you, and as a next step if you could have a look at #553 and do the same for the other PR's as well it would be awesome !

dariuszp:
I just got notification and I was asking myself - why? Then I notice that I was asking about the progress around 2015 :-) Good job! Still I moved away from this project long ago. Parsedown is not safe and should not be used on production.

taylorotwell:
@erusev you are a hero 👍

embluk:
Thank you for merging this update :+1:

PR #499, opened 2017-05-09
Fix hhvm build failure

aidantwoods:
use dist: trusty, on recommendation of https://github.com/travis-ci/travis-ci/issues/7712

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,5 @@ matrix:
   fast_finish: true
   allow_failures:
     - php: nightly
+
+sudo: false


aidantwoods:
Nevermind, I guess

aidantwoods:
@erusev reopening because HHVM is no longer supported on Precise

aidantwoods:
phpunit that is pulled by the current travis hhvm build is broken, see https://github.com/travis-ci/travis-ci/issues/7712#issuecomment-300501839

This PR also moves all builds over to Trusty, and special cases PHP 5.3 on Precise (PHP 5.3 not supported on Trusty with travis)

aidantwoods:
@erusev any chance on merging this so tests on my other PR can pass? 😉 

No major changes here – basically just some config to work around the fact that no single distro on Travis supports both HHVM and PHP 5.3.

And also to work around some bugs with the stock (old version of) phpunit from that HHVM travis image.

More info on reasoning for both of those changes in the travis issue linked above.

PhrozenByte:
I'm not sure what is going on here since I can run Travis for [`picocms/Pico`](https://github.com/picocms/Pico) with HHVM and PHP 5.3 without any problems (see my [`.travis.yml`](https://github.com/picocms/Pico/blob/pico-1.1/.travis.yml)), see e.g. https://travis-ci.org/picocms/Pico/builds/231991944 😕 ❔ 

The problem might be related to the missing `sudo: false` - I can't see any reason why Parsedown requires a Travis instance with `sudo`... `sudo: false` is recommended anyway, see https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments

aidantwoods:
I'm not sure on the internals of Travis, but possible we have different HHVM builds? (see the note at the top of https://travis-ci.org/erusev/parsedown/jobs/230820765)

In any case, going forward we should not continue using precise for HHVM and expect it to work: https://github.com/travis-ci/travis-ci/issues/7712#issuecomment-300528588

> HHVM team has ceased supporting Precise a while back, so I very much recommend moving on.

PhrozenByte:
> I'm not sure on the internals of Travis, but possible we have different HHVM builds?

Yes, Parsedown is using a fully virtualized, `sudo`-enabled build environment, what is discouraged since 2015 for no apparent reason. Simply add `sudo: false` to Parsedown's `.travis.yml` and everything will work again.

> In any case, going forward we should not continue using precise for HHVM

In any case you shouldn't add a explicit dependency to `trusty` when this isn't necessary. This will definitely cause problems later (in terms of "in some years, when nobody remembers why this unnecessary dependency has been added"). You should use Travis's default environment whenever possible, they will update the default environment as soon as it is stable enough.

HHVM and PHP 5.3 both work just fine on Travis's default environment!

aidantwoods:
> You should use Travis's default environment whenever possible, they will update the default environment as soon as it is stable enough.

Generally I agree, however

> In any case you shouldn't add a explicit dependency to `trusty` when this isn't necessary.

Owner of https://github.com/travis-ci/travis-ci has advised its explicit inclusion for HHVM builds, so I'm not sure what to say here.

I don't think the error is due to lack of `sudo` either, looks like we're just pulling different builds. Core failure seems to be a dependency issue caused by HHVM (see https://github.com/travis-ci/travis-ci/issues/7712#issuecomment-299769256).

Also see https://github.com/travis-ci/travis-ci/issues/7712#issuecomment-299896504
> Thanks for the report. I think it is best to document this, and fail HHVM builds on Precise.

> HHVM and PHP 5.3 both work just fine on Travis's default environment!

Even if they work fine now with `sudo: false`, I'd fully expect that to change when the image for the build config is updated.

---

IMO best action here is cope with an awkward config for now and remove the explicit dependence when Travis defaults to trusty.

PhrozenByte:
> I don't think the error is due to lack of sudo either

It is. Container-based repos don't need to run `apt-get` (they can't, even if they wanted to), so there can't be any `apt-get`-related issues.

> Even if they work fine now with sudo: false, I'd fully expect that to change when the image for the build config is updated.

It's the point of the container-based images that they won't get updated when the update breaks things...

> IMO best action here is cope with an awkward config for now and remove the explicit dependence when Travis defaults to trusty.

👎 

Adding `sudo: false` matches 100% the setup which is explicitly recommended by Travis (and the default since 2015!) and works just fine. This overly complex solution that will definitely break things later is simply not necessary.

aidantwoods:
> It's the point of the container-based images that they won't get updated when the update breaks things...

The builds literally started breaking as the result of an update to the default on May 4th, with no changes to the travis config.

https://travis-ci.org/erusev/parsedown/builds/228402377?utm_source=github_status&utm_medium=notification

https://travis-ci.org/erusev/parsedown/builds/229465103?utm_source=github_status&utm_medium=notification

PhrozenByte:
> The builds literally started breaking as the result of an update to the default

Parsedown is **not** using container-based builds as recommended since 2015 (what is also the default for repos created later). Again, refer to https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments

Parsedown is using fully-virtualized, `sudo`-enabled builds which expressly states that they **will break** on updates, due to the fact that they don't represent a guarantee-to-work snapshot of Travis's build environment, but rather a virtual machine with just a ordinary installation of Ubuntu Precise whose supported ended last month (and that's the reason why HHVM dropped support for precise). Container-based builds use a fixed and guarantee-to-work snapshot of Ubuntu Precise before its deprecation, thus they still work as expected. They simply didn't (and will not) receive any updates since this date, but this is also true for the fully-virtualized environments, the latter just break because they **try** to update themselves.

I'm not sure why we're seriously discussing about this, simply **test it**... 😕

PhrozenByte:
:tada:

Thanks @aidantwoods! 👍 😃

aidantwoods:
I guess when Travis updates this particular image, it'll *hopefully* also move to trusty for newer HHVM builds.

PhrozenByte:
Yes, this is planned for early Q3 2017, see https://blog.travis-ci.com/2017-04-17-precise-EOL

> We’re focusing on making the switch to our default stable image from Ubuntu Precise 12.04 to Ubuntu Trusty 14.04 in early Q3, sometime in July.

> For those using the default `sudo: false` container-based Precise 12.04 images: everything will continue as it is.

aidantwoods:
Good, good. I suppose the only issue I can foresee is the current lack of support for PHP 5.3 in trusty. I guess we can cross that bridge when Travis switches everything over though.

aidantwoods:
@erusev thanks!

You should be able to restart the builds on any PRs that failed due to this.

(e.g. https://travis-ci.org/erusev/parsedown/builds/230820758?utm_source=github_status&utm_medium=notification 😉 )

aidantwoods:
cc: @PhrozenByte 

> It's the point of the container-based images that they won't get updated when the update breaks things...

https://travis-ci.org/erusev/parsedown/jobs/242557650
screen shot 2017-06-13 at 20 36 58

😉

PhrozenByte:
Yep, encountered this yesterday, too. 😒 I've absolutely no idea why this happened, because it contradicts the whole idea of the container-based infrastructure... 😕 Nevertheless, I was wrong, you were right, my fault, sorry for the inconvenience.

I suggest either using `dist: trusty` for HHVM and HHVM-nightly (which can be re-added), or using `dist: trusty` by default and `dist: precise` for PHP 5.3 (see https://github.com/picocms/Pico/commit/66916f62bcdc9bd7dee9aace312fb25771c50475 for the latter). Both solutions are like choosing between the devil and the deep blue sea... 😒

aidantwoods:
> Yep, encountered this yesterday, too. 😒 I've absolutely no idea why this happened, because it contradicts the whole idea of the container-based infrastructure... 😕 Nevertheless, I was wrong, you were right, my fault, sorry for the inconvenience.

No worries – to be honest I have no idea why it's like this either (only reason I was pushing for the other solution was based on what I was told by the guys from Travis). Keeping containers working seems like the saner solution IMO!

PHP 5.3 won't be supported on `trusty` going forward, so restricting it to `precise` seems not too insane an option. As for the `trusty` default, I guess we can just open a bug in this repo to remove it once that build reaches stable within travis?

PhrozenByte:
> PHP 5.3 won't be supported on trusty going forward, so restricting it to precise seems not too insane an option. As for the trusty default, I guess we can just open a bug in this repo to remove it once that build reaches stable within travis?

I agree, good idea! :+1:

PR #502, opened 2017-05-10
Ignore composer.lock and vendor

lucasmichot:
Ignore composer.lock and vendor

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+/vendor
+composer.json
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,2 @@
 /vendor
-composer.json
+composer.lock


hkdobrev:
From Composer documentation:

> #### Commit Your composer.lock File to Version Control
> 
> Committing this file to VC is important because it will cause anyone who sets up the project to use the exact same versions of the dependencies that you are using. Your CI server, production machines, other developers in your team, everything and everyone runs on the same dependencies, which mitigates the potential for bugs affecting only some parts of the deployments. Even if you develop alone, in six months when reinstalling the project you can feel confident the dependencies installed are still working even if your dependencies released many new versions since then.

https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control

I'm personally a proponent of committing composer.lock to version control. However, if it's not committed, I agree it should be ignored as it'd cause an untracked file for everyone trying to contribute and doing `composer install`.

As for the `vendor/` directory, there are no dependencies at the moment, but it should still be ignored. Composer also adds more files for autoloading and `installed.json` there as well.

lucasmichot:
This is a package (that does not have dependencies), not a project.
So composer.lock and vendor can be ignored

.gitignore
```diff
@@ -0,0 +1,2 @@
+/vendor
+composer.json
```
PhrozenByte: Definitely not what you want...

hkdobrev:
@lucasmichot From the Composer docs about libraries:

> #### Lock file
> 
> For your library you may commit the `composer.lock` file if you want to. This can help your team to always test against the same dependency versions. However, this lock file will not have any effect on other projects that depend on it. It only has an effect on the main project

lucasmichot:
Let's go on commiting them then

hkdobrev:
@lucasmichot To clarify, I wasn't against this PR. I still think `vendor/` should be ignored and `composer.lock` should either be committed or ignored. At the moment it's lingering in between which is not nice. But this is up to @erusev to decide. Personally, I'd like the PR to be re-opened so this could be explicitly communicated and decided.

cebe:
vendor should be ignored, there is not sense in committing it.

lucasmichot:
I let you guys open a PR for vendor if you feel so

aidantwoods:
> At the moment it's lingering in between which is not nice. But this is up to @erusev to decide. Personally, I'd like the PR to be re-opened so this could be explicitly communicated and decided.

Ping, +1 to reopen

> I let you guys open a PR for `vendor` if you feel so

I'll open another if necessary, though would like @lucasmichot to get credit – since he brought the issue up

---

(Hi there @lucasmichot! Small world 😉)

lucasmichot:
Hé hé @aidantwoods 😀
Up to you @erusev

I believe that if composer.json is strict enough, composer.lock is unecessary

Issue #507, opened 2017-05-26
Cap drop syntax with Parsedown


Flerex:
Hello,

I'm trying to create a new syntax to format paragraphs. What I'm trying to achieve is to have a paragraph's first letter as a cap drop. The syntax I'd expect the user to use is the following, where, of all the accents, only the first one would make the E a cap-drop.

> ^Et rerum sint laudantium eaque veritatis sunt id. Eos vero architecto molestiae facere blanditiis. Consectetur iure error et fugit quasi. Quaerat maxime tenetur eu^m possimus sunt ea maxime. Inventore impedit similique perferendis illum earum ^nam magnam rerum. Qui ut non voluptatem.

I tried to accomplish this following the "color" code of the Extension tutorial, but the following code would work with all the ^, even if they don't strictly start a paragraph.

```
public function inlineDropCap($excerpt)
{

	if (preg_match('/^\^(.{1})/', $excerpt['text'], $matches)) {
		return [
			'extent'  => strlen($matches[0]),
			'element' => [
				'name' => 'span',
				'text' => $matches[1],
				'attributes' => [
					'class' => 'drop-cap',
				],
			],
		];
	}
	
}
```

Can anyone help me with this?

Thank you in advance.

aidantwoods:
Have you added `^` as a new marker near the top of the Parsedown file?

aidantwoods:
Also, this might be better implemented as a block structure if you only want it to apply at the start of a paragraph

Flerex:
I just tried to change inline to block in the function name and modify a little the function. It is working now, but I found some weird behavior that I guess it's caused because the code is not correctly modified.

If the accent is at the begining of more than one paragraphs. Every cap drop is shown only in the first paragraph with a cap drop. Like this:

```
## Lorem Ipsum

^Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.

[...]

PR #511, opened 2017-06-13
Allow element to have no name

aidantwoods:
Currently, Parsedown does not permit text and elements to go side by side in its output structure. Sometimes this is useful, e.g. #507 

Changes allow the element name either as `null` or undefined altogether to mean that only text should be output.


aidantwoods:
Once again, HHVM failures upstream (the rest of the builds pass).

PR #512, opened 2017-06-14
Fix hhvm

aidantwoods:
Fix HHVM (for real this time 😉)

Upon merge, I shall open a bug/PR in this repo to remove the line `dist: trusty`, which should be removed once Travis updates their stable container to trusty.

---

I've also tacked on `hhvm-nightly` as a build stage, since that is supported in trusty (I think?) – which we'll now be using.

cc: @PhrozenByte

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,16 @@ php:
   - 5.6
   - 5.5
   - 5.4
-  - 5.3
   - hhvm
   - nightly
 
 matrix:
+  include:
+    - php: 5.3
+      dist: precise
   fast_finish: true
   allow_failures:
     - php: nightly
 
+dist: trusty
 sudo: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,14 +6,22 @@ php:
   - 5.6
   - 5.5
   - 5.4
-  - hhvm
   - nightly
-  - hhvm-nightly
 
 matrix:
   include:
     - php: 5.3
       dist: precise
+    - php: hhvm
+      install:
+        - composer require phpunit/phpunit --dev --no-update --no-interaction
+        - travis_retry composer install --prefer-dist --no-interaction --no-suggest
+      script: vendor/bin/phpunit
+    - php: hhvm-nightly
+      install:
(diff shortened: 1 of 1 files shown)


aidantwoods:
I'll also add another bug to remove that special case I just added (this special case is following advice from Travis – to update phpunit manually until they include a working version in hhvm trusty).

https://github.com/travis-ci/travis-ci/issues/7712#issuecomment-300509886
> [...] If it is faulty, I suggest updating it yourself until the image is updated.

PhrozenByte:
IMO the appropriate version of `phpunit` should be added to `require-dev` in Parsedown's `composer.json` and Parsedown should always use the self-provided `phpunit` cli, not just for HHVM. However, doing so is controversial, although they officially [mention this option](https://phpunit.de/manual/current/en/installation.html#installation.composer).

aidantwoods:
> IMO the appropriate version of `phpunit` should be added to `require-dev` in Parsedown's `composer.json` and Parsedown should always use the self-provided `phpunit` cli, not just for HHVM. However, doing so is controversial, although they officially mention this option.

I also hold this opinion, though perhaps this belongs in a separate PR? 😉 

I figure having it downloaded as part of the travis HHVM installation most honours Parsedown's current setup (though we could of course have this change, as it would be unneeded if Parsedown later changes to put `phpunit` in `require-dev`)

PhrozenByte:
> I also hold this opinion, though perhaps this belongs in a separate PR? 😉

I think it's quite reasonable to fix both problems in one PR, especially since you'll have to address the issue anyway ("all or nothing" 😆)

aidantwoods:
I'll defer this to @erusev in that case. @erusev would you like me to add
`phpunit` into `require-dev`?

aidantwoods:
@erusev any word on your preference here?

If you don't want to add `phpunit` into `require-dev` then this PR is good to go as is, as far as fixing the tests is concerned.

If you do, then I can add it in and slim down some of the installation instructions in the tests.

gabriel-caruso:
@aidantwoods `Travis CI` was already rewritten in #539. You can close this PR

PR #513, opened 2017-06-19
Added tests for unicode content and updated parser to manage utf-8

Daniel-KM:
The parser is not unicode compliant: see the test https://github.com/erusev/parsedown/pull/513/commits/cc412e41e31473c85c7cad8ee19003e577864dd1#diff-8b846e29de17835941c7735362aefedd
This patch fixes it.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -21,6 +21,20 @@ class Parsedown
 
     # ~
 
+    function __construct()
+    {
+        if (extension_loaded('mbstring')) {
+            mb_regex_encoding('UTF-8');
+            mb_internal_encoding('UTF-8');
+        } elseif (extension_loaded('iconv')) {
+            iconv_set_encoding('input_encoding', 'UTF-8');
+            iconv_set_encoding('internal_encoding', 'UTF-8');
+            iconv_set_encoding('output_encoding', 'UTF-8');
+        }
+    }
+
+    # ~
+
     function text($text)
     {
         # make sure no definitions are set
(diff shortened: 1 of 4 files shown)


Daniel-KM:
There are small changes to do, but everywhere in the parser, so this is based on the last main update of the tool made by @aidantwoods.

NightYoshi370:
Does this have a chance of ever getting implemented?

hkdobrev (changes requested):

aidantwoods:
@NightYoshi370 
> Does this have a chance of ever getting implemented?

I don't see why not :)

---

Couple of comments related to the changes:

Since this is based on #514, it's probably best to get that rebased on master and merged before tackling this one.

I think that in pretty much all cases getting utf-8 compatibility will just be a case of swapping the byte string functions for the `mbstring` ones (as this PR effectively does). This since we actually care about characters in all cases I can recall, and not bytes – bytes just happen to work for common Latin chars.

Given the discussion around #561 and related, since we already use the `mbstring` extension currently, we should just state that dependency (and now do). Hence we should just assume that the `mbstring` functions exist, and not implement our own compatibility layer. If someone really can't install `mbstring` then they will require an extension to implement a polyfill – they are best positioned to choose a polyfill to do this knowing their specific requirements. Alternatively (and preferably), they can just install `mbstring` :) (a polyfill is really only if they have no control over the PHP installation and just have to work with what they have).

aidantwoods:
I've now resolved the merge conflicts in and merged the pre-requisite PR #514.

PR #489, opened 2017-03-25
Htmlblocks

aidantwoods:
Fix https://github.com/erusev/parsedown/issues/488

PR #514, opened 2017-06-21
Added tests for consistency when a markdown follows a markup without blank line

### LINKED ISSUE #488: Invalid parsing syntax in HTML tag
I have following text:

```
<div class="alert alert-warning">
**bold**

# h1
## h2

> quote1
> quote2

* uno
* due

1. one
2. two

[foo.bar](http://foo.bar)
</div>
```

And parer won't parse syntax into `div` wrapper.

Daniel-KM:
When a markdown follows a markup, sometime it is processed, sometime not. The difference occurs when there are one or two markups followed by a markdown. See https://github.com/erusev/parsedown-extra/issues/44 too for markdown extra.

Fixes #488


Daniel-KM:
Phpunit failed, but this is the purpose of this consistency test...

aidantwoods:
Nice spot!

IMO the tests should be positive for **not** processing markdown immediately after HTML markup (since that follows behaviour expected by CommonMark).

Daniel-KM:
I'll commit the inverted test, but perhaps @erusev has a specific point of view. For me, it doesn't matter, since I use Markdown Extra, and I need this check here. 

@erusev, about Markdown Extra, there were no new release since two years, and I'm waiting 1.7.0 of parsedown to fix last issues. Do you plan something or, just missing time?

aidantwoods:
@Daniel-KM if you push those inverted tests, I'll target https://github.com/erusev/parsedown/pull/489 at your branch, so these tests can pass (hopefully).

Daniel-KM:
@aidantwoods Here are the new tests.

aidantwoods:
I've made a PR to your branch, once that's merged you can add the text "Fixes #488" to your the initial description here, so that you can trigger that issue to auto-close (since that was the aim of my PR when targeted at this branch – which will now be part of this one).

aidantwoods:
To say a bit about the changes contained in this PR:

Parsedown currently has a concept of HTML block depth, and will stop processing markdown while the markdown is inside HTML according to the tracked depth.

The changes here essentially swap this model for the CommonMark approach which says (roughly speaking) that a HTML block is a line that starts with a HTML tag and that keeps being a HTML block until the next blank line. Markdown should not be processed unless there is at least one line break between the HTML block and the markdown to be processed. This means that markdown can be processed inside HTML *tags* provided there is a sufficient gap, but acts to limit accidental processing of actual HTML.

One test case in Parsedown's test suite related to these changes failed because it disagreed with the result for the same input to the CommonMark reference parser, this was changed to match the CommonMark result, and so now passes.
All other tests are additive, and have been kindly added by @Daniel-KM.

PR #520, opened 2017-07-27
Add support - [x] AND - [ ]

idreamsoft:
Add support - [x] AND - [ ]

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -515,15 +515,17 @@ class Parsedown
                 ),
             );
 
-            if($name === 'ol') 
+            if($name === 'ol')
             {
                 $listStart = stristr($matches[0], '.', true);
-                
+
                 if($listStart !== '1')
                 {
                     $Block['element']['attributes'] = array('start' => $listStart);
                 }
             }
+
+            $this->checkbox($matches[2],$attributes);
 
             $Block['li'] = array(
                 'name' => 'li',
@@ -532,7 +534,7 @@ class Parsedown
                     $matches[2],
                 ),
             );
-
+            $attributes && $Block['li']['attributes'] = $attributes;
             $Block['element']['text'] []= & $Block['li'];
 
             return $Block;
@@ -554,6 +556,8 @@ class Parsedown
 
             $text = isset($matches[1]) ? $matches[1] : '';
 
+            $this->checkbox($text,$attributes);
+
             $Block['li'] = array(
                 'name' => 'li',
                 'handler' => 'li',
@@ -561,6 +565,7 @@ class Parsedown
                     $text,
                 ),
             );
+            $attributes && $Block['li']['attributes'] = $attributes;
 
             $Block['element']['text'] []= & $Block['li'];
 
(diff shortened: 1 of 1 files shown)


Parsedown.php
```diff
@@ -1453,8 +1458,19 @@ protected function elements(array $Elements)
...
     }
 
+    #
+    # - [x] AND - [ ]
+    protected function checkbox(&$text,&$attributes)
+    {
+        if(strpos($text,'[x]')!==false||strpos($text,'[ ]')!==false){
+            $attributes = array("style"=>"list-style: none;");
+            $text = str_replace(array('[x]','[ ]'), array(
+                '<input type="checkbox" checked="true" disabled="true">',
```
aidantwoods: This should be refactored into the internal way of representing an element so that it doesn't break when #495 is merged (see #507 for an example).

Parsedown.php
```diff
@@ -1453,8 +1458,19 @@ protected function elements(array $Elements)
         return $markup;
     }
 
+    #
+    # - [x] AND - [ ]
+    protected function checkbox(&$text,&$attributes)
+    {
+        if(strpos($text,'[x]')!==false||strpos($text,'[ ]')!==false){
+            $attributes = array("style"=>"list-style: none;");
```
aidantwoods: How does this affect nested (non checkbox style) lists

Parsedown.php
```diff
@@ -1453,8 +1458,19 @@ protected function elements(array $Elements)
         return $markup;
     }
 
+    #
+    # - [x] AND - [ ]
+    protected function checkbox(&$text,&$attributes)
```
aidantwoods: Perhaps have this method build the entire `li` and return that, to avoid returning multiple pieces of data/using references?

Parsedown.php
```diff
@@ -515,24 +515,26 @@ protected function blockList($Line)
...
+
             $Block['li'] = array(
                 'name' => 'li',
                 'handler' => 'li',
                 'text' => array(
                     $matches[2],
                 ),
             );
-
+            $attributes && $Block['li']['attributes'] = $attributes;
```
aidantwoods: Can this be refactored into an `if` for readability?

Parsedown.php
```diff
@@ -554,13 +556,16 @@ protected function blockListContinue($Line, array $Block)
...
+            $this->checkbox($text,$attributes);
+
             $Block['li'] = array(
                 'name' => 'li',
                 'handler' => 'li',
                 'text' => array(
                     $text,
                 ),
             );
+            $attributes && $Block['li']['attributes'] = $attributes;
```
aidantwoods: Can this be refactored into an `if` for readability?

Parsedown.php
aidantwoods: Also Parsedown has no precedent of inline styles thus far – this could break in applications that deploy a CSP, perhaps a non `li` could be used instead for the entire structure?

aidantwoods: The GFM spec does not advise using an inline style here (see:  https://github.github.com/gfm/#task-list-items-extension-)

idreamsoft:
OK
Thank you for pointing out the inadequacies
I'll change it again
Your code is very strict :)

aidantwoods:
> Your code is very strict :)

Aha 😆 Most of this should be taken as my opinion/my guided opinion based on conventions already used in the code elsewhere. Final decision will be up to @erusev, it's his library ;)

tovic:
IMO, I think it is better if created in a separated extensions, as **Parsedown** has **ParsedownExtra**, then maybe we can just have another extension called **ParsedownGFM** (for example). Stay on the common features following the [CommonMark](http://spec.commonmark.org) spec as the core. Keep it simple.

aidantwoods:
@tovic isn't one of Parsedown's features in the readme that it is Github flavoured? If anything, this is a bugfix? 😉

PR #525, opened 2017-08-20
Properly support fenced code block infostring

### LINKED ISSUE #524: incorrect render for c#
can not render c#, only csharp
```
~~~ c#
this is c# code
~~~
```

this is ok.
```
~~~ csharp
this is c# code
~~~
```

aidantwoods:
Reference: http://spec.commonmark.org/0.28/#info-string

---

Fixes #524

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -396,7 +396,7 @@ class Parsedown
 
     protected function blockFencedCode($Line)
     {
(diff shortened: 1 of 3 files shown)


aidantwoods:
As always, #512 will need merging for tests to pass here

PR #533, opened 2017-10-12
Fix identification of closing tags.

tadeas:
A line like this:
</div>  

Was not identified as a closing tag.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -740,7 +740,7 @@ class Parsedown
             $Block['depth'] ++;
         }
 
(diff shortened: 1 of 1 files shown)


aidantwoods:
We've changed the way HTML blocks are dealt with now, so it works more like the way that CommonMark treats HTML – i.e. we no longer try to determine whether we are within HTML tags, rather we just stop processing markdown that is part of a HTML block (as defined by CommonMark).

Hopefully this should help cause less problems like this, where closing tags could be missed.

Some details in #514

PR #538, opened 2017-11-11
added headerless table

jheimbach:
if header content is empty, the html isnot included


erusev:
That's interesting, is there a markdown implementation where tables work like this? Thanks

cebe:
https://johnmacfarlane.net/babelmark2/?normalize=1&text=%7C+%7C+%7C%0A%7C+--------+%7C+--------+%7C%0A%7C+cell+1.1+%7C+cell+1.2+%7C%0A%7C+cell+2.1+%7C+cell+2.2+%7C

pandoc is the only one that renders no header as far as I see:

 https://johnmacfarlane.net/babelmark2/?normalize=1&text=%7C%7C%7C%0A%7C+--------+%7C+--------+%7C%0A%7C+cell+1.1+%7C+cell+1.2+%7C%0A%7C+cell+2.1+%7C+cell+2.2+%7C

Github renders like this:

|||
| -------- | -------- |
| cell 1.1 | cell 1.2 |
| cell 2.1 | cell 2.2 |

aidantwoods:
How about something like

```markdown
| --- | --- |
| abc | def |
```

For a headerless table?

JHeimbach:
@aidantwoods that looks (and feels) cleaner than my proposal, I like it.

aidantwoods:
I opened a new PR for the suggested syntax above (https://github.com/erusev/parsedown/pull/580).

A couple things to note about
```markdown
| --- | --- |
| abc | def |
```
Pros:
* No need to write empty structure delimiters
* Very analogous to an already supported syntax in GitHub flavoured markdown for tables with no body:
  ```markdown
  | abc | def |
  | --- | --- |
  ```
* Syntax is currently an invalid table in GitHub favoured markdown, so will not collide with any sort of already valid table

Cons:
* The failure mode for text that uses a different parser is likely to be no table rendering at all (whereas the other variant proposed here would render a table with empty whitespace in the header as a fallback).
* No one currently seems to support this variation: [Babelmark2](https://johnmacfarlane.net/babelmark2/?normalize=1&text=%7C+--------+%7C+--------+%7C%0A%7C+cell+1.1+%7C+cell+1.2+%7C%0A%7C+cell+2.1+%7C+cell+2.2+%7C)
* Deviates from GitHub flavoured markdown spec

I proposed this as a syntax to the GitHub flavoured markdown spec (https://github.com/github/cmark/issues/91), though the repo seems not to be as active as it once was.

PR #539, opened 2017-11-11
Use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase

gabriel-caruso:
I use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCases. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

I just need to require PHPUnit version [`^4.8.35`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-4.8.md#4835---2017-02-06), that support this namespace.

I also change `Travis CI` to use the installed `PHPUnit` version, not the global one.


erusev:
> I also change Travis CI to use the installed PHPUnit version, not the global one.

What's the problem with the global one?

gabriel-caruso:
@erusev If you take a look on [this](https://travis-ci.org/erusev/parsedown/builds/300469898?utm_source=github_status&utm_medium=notification) Travis CI fails, it tries to use a global PHPUnit, version 5 or 6 that fails.

erusev:
Do you know why the global version fails? Can we resolve that fail more elegantly? Thanks!

gabriel-caruso:
@erusev Failed because it tries to use the latest version of `PHPUnit`, current `PHPUnit 6`. Want me to support it? And what about `PHP 5.3`? Should we `allow_failures`?

erusev:
I think we can allow failures for 5.3.

Can we address the failures without introducing dev pages and making the build process more complicated?

Thanks.

gabriel-caruso:
Sorry, I didn't get it: `dev-pages`?

PhrozenByte:
Adding `phpunit` to composer's `require-dev` and using the `phpunit` script provided by composer is AFAIK the common solution - precisely because you usually want to explicitly use a specific version of `phpunit`.

gabriel-caruso:
A @PhrozenByte said. If you want, [here is how Laravel handles it](https://github.com/laravel/framework/blob/5.5/.travis.yml).

erusev:
@PhrozenByte Makes sense, do you have recommendations as to which PHPUnit version we should use and also, could you have a look at the proposed changes and let us know if you have any suggestions? Thanks

PhrozenByte:
Depends. The current stable PHPUnit 6 requires PHP 7.0+, oldstable PHPUnit 5 requires PHP 5.6+. PHPUnit 4 is already end-of-life (requires PHP 5.3+). So we have basically two options:

1. PHPUnit is not security-related, thus it's IMO feasible to use PHPUnit 4 (as @gabriel-caruso currently suggests it)
2. Don't test Parsedown for PHP 5.3, 5.4, 5.5 (basic syntax checking should be possible though). However, I strongly recommend **not** to drop support of PHP 5.3.

gabriel-caruso:
@erusev So, as the PR is, is there something else to do?

erusev:
@gabriel-caruso 

Do you have an opinion on what @PhrozenByte says about `self-update`?

Also, what problem does `sudo: false` solve?

image

Thanks!

gabriel-caruso:
I’m going to remove `sudo` and `self-update` and run the tests

aidantwoods:
@erusev 

> It's not going anywhere, but it seems like there's no elegant way to make TravisCI run tests on 5.3.

PHP 5.3 failures are solved in #512 (which is now out of date in some places due to updates in travis). You just need to explicitly run 5.3 tests on precise.

.travis.yml
```diff
@@ -14,5 +14,10 @@ matrix:
   fast_finish: true
   allow_failures:
     - php: nightly
+    - php: 5.3
 
-sudo: false
\ No newline at end of file
+before_script:
+  - composer install --prefer-dist --no-interaction --no-progress
```
aidantwoods: Consider adding `travis_retry` prior to this command in-case of a temporary network issue

(i.e. to read: `travis_retry composer install --prefer-dist --no-interaction --no-progress`)

gabriel-caruso:
As we've done all this job, let's support `PHPUnit` versions 5 and 6?

erusev:
Awesome, I'm merging this

PR #541, opened 2017-11-19
Removing mbstring dependency

skratte:
The dependency to mbstring in php7 could be removed with a fallback to strlen.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -141,7 +141,14 @@ class Parsedown
 
                 foreach ($parts as $part)
                 {
-                    $shortage = 4 - mb_strlen($line, 'utf-8') % 4;
+                    if(function_exists('mb_strlen'))
+                    {
+                        $shortage = 4 - mb_strlen($line, 'utf8') % 4;
+                    }
+                    else
+                    {
+                        $shortage = 4 - strlen($line) % 4;
+                    }
 
                     $line .= str_repeat(' ', $shortage);
                     $line .= $part;
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -141,11 +141,11 @@ class Parsedown
 
                 foreach ($parts as $part)
                 {
-                    if(function_exists(mb_strlen))
+                    if(function_exists('mb_strlen'))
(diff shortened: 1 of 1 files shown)


cebe:
Could you provide a source for that? 

According to the documentation your change is wrong:

>  strlen() returns the number of bytes rather than the number of characters in a string. 
> http://php.net/manual/en/function.strlen.php

PhrozenByte:
How about using https://github.com/symfony/polyfill-mbstring?

skratte:
You are right. I was too careless and didn't see the differences in the results. 

I did a little more research on this and the code could be like this:

<pre><code>
                foreach ($parts as $part)
                {
                    if(function_exists('mb_strlen'))
                    {   
                        $shortage = 4 - mb_strlen($line, 'UTF-8') % 4;
                    }
                    elseif(function_exists('iconv_strlen'))
                    {   
                        $shortage = 4 - @iconv_strlen($line, 'UTF-8') % 4;
                    }
                    else
                    {
                        $shortage = 4 - preg_match_all("/.{1}/us",$line) % 4;
                    }

                    $line .= str_repeat(' ', $shortage);
                    $line .= $part;
                }
</code></pre>

This way it will fallback from non-default extension to default extension to core extension. 

As stated in a comment to iconv_strlen, it will fail with badly formatted sequences. Because of this i had to silence iconv_strlen. 

With non of the solutions i get the expected results on bad sequences. To test bad sequences i used this: http://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-test.txt

All in all, the code is not pretty but it works as it should.

PhrozenByte:
Just one small addition:

If you're developing a end-user application with Parsedown, and if you don't want to require your users to install `ext-mbstring`, simply add the following to your `composer.json`:

```json
{
    "require": {
        "erusev/parsedown": "^1.7.1",
        "symfony/polyfill-mbstring": "^1.7"
    },
    "config": {
        "platform": { "ext-mbstring": "7" }
    }
}
```

PR #544, opened 2017-11-28
Skip Vue.js compilation within <code> blocks

AlexMordred:


aidantwoods:
Adding this attribute would go against CommonMark spec – but you wanted to add this for your own use then this is probably best implemented via an extension :)

PR #548, opened 2017-12-31
Update license year

m1guelpf:
TO DO:

- [x] Wait to the next year (100% done, depends on timezone)
- [ ] Merge!

:octocat:


erusev:
Ha, that's fast, can we keep the original year and do sth like 2013 - 2018, also, the commit has an unnecessary change on the last line.

m1guelpf:
@erusev Updated! The last line change seems to be a newline auto-added by GitHub's UI...

PR #556, opened 2018-02-28
Adjust version constant to 1.7.0

stefandoorn:


PhrozenByte:
Since merging this would require tagging a new release the version constant should rather be `1.7.1` or something :wink:

stefandoorn:
Haha indeed :-)

scottchiefbaker:
This does need to get addressed. The version constant has been out of whack for quite a while now. Even when 1.6.1 was released it's always said 1.6.0.

aidantwoods:
I think it should be possible to parse `git describe --tags` in Travis to detect if it is a release commit and compare against the hard coded version so the build would at least fail if the number doesn't get updated. It would be nice to get this to auto-update as part of the release process too really

PR #557, opened 2018-02-28
Talk about safe mode in the README

aidantwoods:


PhrozenByte:
* The `README.md` should also cover how to just escape HTML without XSS protection
* We shouldn't suggest using Parsedown's safe mode *together* with e.g. HTMLPurifier because this makes not much sense. One should *either* use Parsedown's safe mode *or* HTMLPurifier - they do different things and cover different use cases. Safe mode makes the output safe due to disallowing custom HTML and closing attack vectors caused by pure Markdown. HTMLPurifier makes the output safe due to blacklisting bad contents while still allowing custom HTML (as long as the custom HTML is safe)
* CSP is IMO not really related to Parsedown, but rather some general security topic. I suggest removing this sentence

aidantwoods:
> We shouldn't suggest using Parsedown's safe mode together with e.g. HTMLPurifier because this makes not much sense. One should either use Parsedown's safe mode or HTMLPurifier - they do different things and cover different use cases. Safe mode makes the output safe due to disallowing custom HTML and closing attack vectors caused by pure Markdown. HTMLPurifier makes the output safe due to blacklisting bad contents while still allowing custom HTML (as long as the custom HTML is safe)

Yes, this is probably worth clarifying :)

> CSP is IMO not really related to Parsedown, but rather some general security topic. I suggest removing this sentence

CSP is a browser security feature you can take advantage of protect against XSS (in-fact it'll basically kill it if you set it up well). While safe mode should work, IMO its a good idea to recommend users set up multiple layers of defence anyway :)

README.md
```diff
@@ -38,7 +38,20 @@ More examples in [the wiki](https://github.com/erusev/parsedown/wiki/) and in [t
 
 ### Security
 
-Parsedown does not sanitize the HTML that it generates. When you deal with untrusted content (ex: user comments) you should also use a HTML sanitizer like [HTML Purifier](http://htmlpurifier.org/).
+Parsedown is capable of escaping user-input within the HTML that it generates.
+Additionally Parsedown can attempt to sanitize additional scriping vectors (such
```
drbyte: `scriping` => `scripting`

aidantwoods: Thanks :)

Issue #232, opened 2014-10-06
Rendering ```php+html issues


harikt:
When rendering this source https://github.com/harikt/aurav2book/blob/186fb12a5bfedbb62b45c9fd5d9224f3cc22deb5/manuscript/helpers.md#a-a it is producing wrong output .

See http://ghdocs.herokuapp.com/harikt/aurav2book/186fb12a5bfedbb62b45c9fd5d9224f3cc22deb5/manuscript/helpers.md#a

Tag is like 

```
html+php
<?php
echo $this->a(
    'http://auraphp.com',       // (string) href
    'Aura Project',             // (string) text
    array('id' => 'aura-link')  // (array) optional attributes
);
?>
<a href="http://auraphp.com" id="aura-link">Aura Project</a>
```

Removed the 3 ticks.

erusev:
@harikt 

What part of the output is wrong?

harikt:
Did you noticed the link and the syntax I wrote :) .

erusev:
It'd be best if you could provide a short Markdown text together the expected HTML.

hkdobrev:
I think when there is a `+` character after the "```" characters it doesn't get parsed as a code block, because of the regex used or something similar.

harikt:
This is markdown. I wonder whether you will see the real markdown for this will be rendered by github .

    ```html+php
    <?php
    echo $this->a(
        'http://auraphp.com',       // (string) href
        'Aura Project',             // (string) text
        array('id' => 'aura-link')  // (array) optional attributes
    );
    ?>
    <a href="http://auraphp.com" id="aura-link">Aura Project</a>
    ```

The expected output is 

```php
<?php
echo $this->a(
    'http://auraphp.com',       // (string) href
    'Aura Project',             // (string) text
    array('id' => 'aura-link')  // (array) optional attributes
);
?>
<a href="http://auraphp.com" id="aura-link">Aura Project</a>
```

and what it produce is 

```
```html+php <?php echo $this->a(

'http://auraphp.com',       // (string) href
'Aura Project',             // (string) text
array('id' => 'aura-link')  // (array) optional attributes
); ?>
```

harikt:
Hey @aidantwoods , 

I guess the problem is with the finding of regex . ` ```html+php ` , ie a wild guess.

Eg : This one uses two languages, html and php. So may be if I change to make use of only php this issue will be fixed I guess.

Not a major issue for me. Thanks once again.

harikt:
[...]

PR #558, opened 2018-02-28
Add test case to make sure issue 232 no longer exists

harikt:
Apologies for not sending a failure test in https://github.com/erusev/parsedown/issues/232 . Too late, but hope I will address the same at-least now.

Thank you.


aidantwoods:
Thanks! 😃

harikt:
I don't know why I didn't send a failing test earlier. But your PR #524 actually helped to add one.

PR #566, opened 2018-03-09
Email autolink shouldn't be started by HTML tags

### LINKED ISSUE #565: HTML tags around email addresses produce mangled ouput
I expect input like `<strong>first.last@example.com</strong>` to produce output like `<strong>first.last@example.com</strong>`. Instead, we see output like `<a href="mailto:strong&gt;first.last@example.com&lt;/strong">strong&gt;first.last@example.com&lt;/strong</a>`.

I suggest something like:

```
diff --git a/Parsedown.php b/Parsedown.php
index 87d612a..39a3aed 100644
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1146,6 +1146,10 @@ class Parsedown
         {
             $url = $matches[1];
 
+            if (!filter_var($url, FILTER_VALIDATE_EMAIL)) {
+                return;
+            }
+
             if ( ! isset($matches[2]))
             {
                 $url = 'mailto:' . $url;
```

aidantwoods:
Fixes #565

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1142,10 +1142,10 @@ class Parsedown
 
     protected function inlineEmailTag($Excerpt)
     {
-        $commonMarkEmail = '[a-zA-Z0-9.!#$%&\'*+\/=?^_`{|}~-]++@[a-zA-Z0-9]'
-            .'(?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?'
-            .'(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*'
-        ;
+        $hostnameLabel = '[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?';
+
+        $commonMarkEmail = '[a-zA-Z0-9.!#$%&\'*+\/=?^_`{|}~-]++@'
+            . $hostnameLabel . '(?:\.' . $hostnameLabel . ')*';
 
         if (strpos($Excerpt['text'], '>') !== false
             and preg_match("/^<((mailto:)?$commonMarkEmail)>/i", $Excerpt['text'], $matches)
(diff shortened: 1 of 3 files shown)


aidantwoods:
The regex looks a bit huge, but it isn't very complex (the majority is defining character classes), so I don't think it'll be any (much?) worse than the existing regex complexity (open to performance suggestions though, cc: @PhrozenByte)

I think `filter_var`'s email filter doesn't match the definition that CommonMark references (CommonMark uses the HTML5 definition), but I'm not sure of the exact discrepancies.

Parsedown.php
```diff
@@ -1142,8 +1142,14 @@ protected function inlineCode($Excerpt)
 
     protected function inlineEmailTag($Excerpt)
     {
-        if (strpos($Excerpt['text'], '>') !== false and preg_match('/^<((mailto:)?\S+?@\S+?)>/i', $Excerpt['text'], $matches))
-        {
+        $commonMarkEmail = '[a-zA-Z0-9.!#$%&\'*+\/=?^_`{|}~-]++@[a-zA-Z0-9]'
```
PhrozenByte: Just making the regex a little easier to read:

```php
$hostnameLabel = '[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?';
$commonMarkEmail = '[a-zA-Z0-9.!#$%&\'*+\/=?^_`{|}~-]++@' . $hostnameLabel . '(?:\.' . $hostnameLabel . ')*';
```

aidantwoods:
> The regex was taken from CommonMark, right?

Yup, (though they ripped it straight out the HTML5 spec ;p). The only changes I made were escaping a backslash and a made the "username" part of the email possessive (since that character class is mutually exclusive with `@` I don't think we need backtracking?).

PhrozenByte:
Alright. Yeah, no backtracking necessary there.

PR #574, opened 2018-03-18
Remove legacy escaping

### LINKED ISSUE #573: Continue #572: Problem with BBCode and Markdown
Yeah, I realised. Just that my board members love BBCode more than Markdown, but we like having markdown on the wiki.

Also, uh, another thing I've realised:

`Posted by "DestroyerXL44" id="3297"`

But the &-quot; thing replaces the " (sorry, github formatting)

I've tried `->setMarkupEscaped(false),` but it didn't fix that.

I've also tried `setSafeMode(true),` but neither did that. fix it

For reference, below is my configuration:

```php
	$text = Parsedown::instance()
		->setUrlsLinked(false)		# prevents automatic linking of URLs
		->setMarkupEscaped(false)
		->line($text);
```

aidantwoods:
This uses the newly added conditionally escaped HTML from #569 in combination with text-element adjacency from #511 to remove some special cased exceptions for HTML to circumvent the AST. In doing so we can get rid of the last of the old ways of escaping things.

Before merge I'm going to apply similar tests to those I did in #495 (i.e. a combination of manual testing and seeing if OWASP Zap can find any XSS with safe mode enabled).

cc: @PhrozenByte for review :)

---

Fixes #573

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1189,7 +1189,7 @@ class Parsedown
         if (isset($Excerpt['text'][1]) and in_array($Excerpt['text'][1], $this->specialCharacters))
         {
             return array(
-                'markup' => $Excerpt['text'][1],
+                'element' => array('rawHtml' => $Excerpt['text'][1]),
                 'extent' => 2,
             );
         }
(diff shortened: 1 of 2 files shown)


aidantwoods:
No (actual) XSS issues found by Zap (one false-positive which is given below).

screen shot 2018-03-25 at 20 24 00

PR #576, opened 2018-03-21
Produce AST prior to render

### LINKED ISSUE #338: Run a completion block after text() has processed lines()
I was building an extension and I wanted to run something after the blocks have been parsed by `lines` in `text`, the `lines` method is private (in `Parsedown`), and I wanted to do something after everything has been processed (but within my extension and not within my app).

A `Complete` callback to the `text()` function.

`in Parsedown.php`

```
function text($text)
{
    # make sure no definitions are set
    $this->DefinitionData = array();

    # standardize line breaks
    $text = str_replace(array("\r\n", "\r"), "\n", $text);

    # remove surrounding line breaks
    $text = trim($text, "\n");

    # split text into lines
    $lines = explode("\n", $text);

    # iterate through lines to identify blocks
    $markup = $this->lines($lines);

    # trim line breaks
    $markup = trim($markup, "\n");

    # CALLBACK: I want to run something here <---------------------
    // e.g: $this->completeCallback($markup) <-- Which I could override
    // Do custom processing before sending back to the client application (from an extension)

    return $markup;
}
```

---

Currently, I do something like this (This is within my app, I want to do this within my extension, once all the parsing is complete, I want to do some other stuff).

```
    $parsedown = new ParsedownExtension();
    $html = $parsedown->text($string);
    $html = $parsedown->completeCallback($html);
```

`ParsedownExtension` is a child of `Parsedown`, which I add the `completeCallback` and call it manually.

It would be great to have a function which is called after all processing is done which we could override to do custom things after, the other option is for me to override all the other blocks and count the lines to see where we are, and on the last line I could do whatever I wanted - obviously this is quite alot of work!

Currently, I'm using my solution above and doing this in my app, but If I want to use this in another app or publish it online, I would have to instruct everyone to call the callback (when I would really want my extension to do this - run a callback when everything is finished so I can do any custom processing in my extension before passing the `$markup` back to the client application).

Thoughts or ideas for a workaround? - I have to do it after everything is processed as the `processTag` messes things up when inside a `div` (`markdown="1"`).

### LINKED ISSUE #413: remove markdown
can we get this feature? i want parsedown to remove markdown & convert to plain text

aidantwoods:
An experimental go at refactoring the main to produce a complete AST prior to converting it into a HTML string. This could allow us/extensions to traverse the AST and perform structural transformations to parsing results.
Highly experimental, definitely not finished.

---

Closes #338
Closes #413

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1546,9 +1546,18 @@ class Parsedown
                 $Element['nonNestables'] = array();
             }
 
-            $function = $Element['handler']['function'];
-            $argument = $Element['handler']['argument'];
-            $destination = $Element['handler']['destination'];
+            if (is_string($Element['handler']))
+            {
+                $function = $Element['handler'];
+                $argument = $Element['text'];
+                $destination = 'rawHtml';
+            }
+            else
+            {
+                $function = $Element['handler']['function'];
+                $argument = $Element['handler']['argument'];
+                $destination = $Element['handler']['destination'];
+            }
 
             $Element[$destination] = $this->{$function}($argument, $Element['nonNestables']);
         }
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -143,6 +143,11 @@ class Parsedown
 
     protected function lines(array $lines)
     {
+        return $this->elements($this->linesElements($lines));
+    }
+
+    protected function linesElements(array $lines)
+    {
         $CurrentBlock = null;
 
         foreach ($lines as $line)
@@ -284,7 +289,7 @@ class Parsedown
 
         # ~
 
-        $markup = '';
+        $Elements = array();
 
         foreach ($Blocks as $Block)
         {
@@ -293,15 +298,12 @@ class Parsedown
                 continue;
             }
 
-            $markup .= "\n";
-            $markup .= $this->element($Block['element']);
+            $Elements[] = $Block['element'];
         }
-
-        $markup .= "\n";
 
         # ~
 
-        return $markup;
+        return $Elements;
     }
 
     protected function isBlockContinuable($Type)
@@ -386,7 +388,10 @@ class Parsedown
         if (isset($Line['text'][3]) and $Line['text'][3] === '-' and $Line['text'][2] === '-' and $Line['text'][1] === '!')
         {
             $Block = array(
-                'element' => array('rawHtml' => $Line['body']),
+                'element' => array(
+                    'rawHtml' => $Line['body'],
+                    'autobreak' => true,
+                ),
             );
 
             if (preg_match('/-->$/', $Line['text']))
@@ -776,7 +781,10 @@ class Parsedown
 
             $Block = array(
                 'name' => $matches[1],
-                'element' => array('rawHtml' => $Line['text']),
+                'element' => array(
+                    'rawHtml' => $Line['text'],
+                    'autobreak' => true,
+                ),
             );
 
             return $Block;
@@ -1045,7 +1053,12 @@ class Parsedown
 
     public function line($text, $nonNestables=array())
     {
-        $markup = '';
+        return $this->elements($this->lineElements($text, $nonNestables));
+    }
+
+    protected function lineElements($text, $nonNestables = array())
+    {
+        $Elements = array();
 
         # $excerpt is based on the first occurrence of a marker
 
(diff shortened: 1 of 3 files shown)


aidantwoods:
So far this aims to allow the entire AST to be generated prior to its rendering by separating handler delegation from the structure – integrated AST-handler model the AST is generated in layers and then rendered before generating the next later (i.e. generate, render, recurse via handler).

Previously handlers would be required to return HTML, and would be inserted directly within an element. In order to successfully decouple this process and obtain a meaningful AST end result, handlers therefore need to be able to return AST data.
Previously the handler was also used to generate the HTML for nested elements, since this aims to decouple the handler from the AST structure there are now two new keys in addition to `text` within an element: `elements`, and `element`. The other keys allow for nested structures. The introduction of these special keys for AST structure eliminate the need for use of `elements` as a handler during the rendering process. Now only text should be put in `text` and it no longer serves the dual purpose of "argument to handler". 

The handler is now an array that contains information only relevant to handler delegation, in which the following keys should all be specified: `function`, `argument`, `destination`. `function` is the function which should be called during the handling process, `argument` is the argument that should be given to the function, `destination` is the key to which the result of this call should be written to within the element.

IMO this also majorly disambiguates the handling process, in particular the use of the `text` key to contain: text, elements, an element, and lines in an `li`, ... are a little confusing to say the least :p 

This has caused confusion in the past: e.g. https://github.com/erusev/parsedown/issues/507#issuecomment-304463832
> I just tried your code and it works[...] I'm also trying to understand how you did it but I'm having a hard time[...]

It's definitely worth considering the implications of all these changes before deciding which version this might belong to. This is a bit of an experiment, so there is potentially room for improvement with regard to breaking changes.

aidantwoods:
Also note that (since the handler key is now a different type), we could in theory implement a compatibility layer using `rawHtml` (probably without opting into the safe-mode exception) by writing something like:

```php
    protected function handle(array $Element)
    {
        if (isset($Element['handler']))
        {
            if (!isset($Element['nonNestables']))
            {
                $Element['nonNestables'] = array();
            }

            if (is_string($Element['handler']))
            {
                $function = $Element['handler'];
                $argument = $Element['text'];
                $destination = 'rawHtml';
            }
            else
            {
                $function = $Element['handler']['function'];
                $argument = $Element['handler']['argument'];
                $destination = $Element['handler']['destination'];
            }

            $Element[$destination] = $this->{$function}($argument, $Element['nonNestables']);
        }

        unset($Element['handler']);

        return $Element;
    }
```

Just to give an example of doing this with minimal BC breaks if desired.

Sommerregen:
@aidantwoods Since you switch to an AST structure, is there any chance to register node visitors in order to extend functions or add new functionalities? I think something in the direction of #333 .

aidantwoods:
@Sommerregen would you mind opening a separate issue to track this? I'll try take a look at this when I can, though probably separately to this PR :)

aidantwoods:
cc: @PhrozenByte :)
Any thoughts on this – as far as I can see BC breakage should be very minimal (only breakage I can foresee would be in extensions that utilise custom (their own) handlers and expect safe mode to be enabled by their users – the old way of doing handlers that return HTML now goes via conditional escaping in `rawHtml` as a handler destination. There is a pretty easy routes to fix in this case anyway without using the AST (though of course that is preferable).

PhrozenByte:
I'll definitly look into this, unfortunately my time is very limited right now. I hope I'll find some tme in the next couple of days :smiley:

aidantwoods:
Even though it is now possible to generate the AST prior to rendering – `element` will still delegate work to handlers if they are present. This means it is not required to generate the AST in entirety prior to rendering if we don't need to.
If we do wish to see the entire AST then we only need call one of the added recursive methods on the AST to ensure that all handlers are called. The point here was to separate the two processes, we won't always need it though – in these cases we can (and probably should) delay traversal until render.

aidantwoods:
I think I'd consider this pretty much done now, bar any improvements @PhrozenByte may suggest :)

hamedgasemi200:
I hope you finish this soon, I check this page every day !

sbrl:
Is there a hook I can register against when subclassing Parsedown to grab a hold of that AST? It would be great for things like generating tables of contents (ref: https://github.com/sbrl/Pepperminty-Wiki/issues/155)

PR #589, opened 2018-03-29
Support #hashtag per CommonMark and GFM specs

NathanBaulch:
Please consider better supporting lines that begin with #hashtags.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -90,6 +90,15 @@ class Parsedown
 
     protected $safeMode;
 
+    function setHastagsEnabled($hashtagsEnabled)
+    {
+        $this->hashtagsEnabled = (bool) $hashtagsEnabled;
+
+        return $this;
+    }
+
+    protected $hashtagsEnabled;
+
     protected $safeLinksWhitelist = array(
         'http://',
         'https://',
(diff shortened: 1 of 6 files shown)


Parsedown.php
```diff
@@ -506,6 +506,11 @@ protected function blockHeader($Line)
             }
 
             $text = trim($Line['text'], '#');
+
+            if (!isset($text[0]) or $text[0] !== ' ') {
```
aidantwoods: If this were to be merged, I think in order to give some recourse to users that do not want the space requirement, it would be a good to add a new setter which can turn this behaviour off.

NathanBaulch: Easy enough to add a setting to maintain backward compat.

Parsedown.php
```diff
@@ -84,6 +84,15 @@ function setSafeMode($safeMode)
 
     protected $safeMode;
 
+    function setHastagsEnabled($hashtagsEnabled)
```
aidantwoods: Apologies for not coming up with this earlier – could we call this `setStrictMode` and the respective property `strictMode`.

As suggested [here](https://github.com/erusev/parsedown/issues/265#issuecomment-70016695) I think that it would be good to group these opt-in spec improvements under one setter.

test/ParsedownTest.php
```diff
@@ -32,6 +32,7 @@ protected function initDirs()
     protected function initParsedown()
     {
         $Parsedown = new TestParsedown();
+        $Parsedown->setHastagsEnabled(true);
```
aidantwoods: Could you add separate test cases instead of enabling the mode universally?

aidantwoods:
I've made the requested changes myself so we can get this merged.

Thanks for the PR @NathanBaulch! :)

PR #614, opened 2018-04-12
General and performance tweaks

aidantwoods:
This might be the first PR in a while that *reduces* the size of the codebase very slightly ;p

As measured by a few markbench runs, this PR makes Parsedown about 15-20% faster on the github sample profile, and about 10% faster when run without a profile specified. These being run on PHP 7.2.4, and performance diffs are in comparison to 1.8.0-beta-1.

Its probably worth cherry-picking some of these commits over some others. Adjustments to the `inlineText` function have major performance improvements, v.s. things like swapping regular expressions for loops – which doesn't appear to offer much of an observable benefit on my system (latest PHP version at present).

Adjustments towards regular expressions may be beneficial where Parsedown would previously encounter catastrophic backtracking with certain inputs.

Adjustments to string concatenation syntax appear to offer no observable difference in Parsedown in my testing (despite historic benchmarks from the PHP community apparently indicating a difference).

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1105,7 +1105,7 @@ class Parsedown
     # ~
     #
 
-    public function line($text, $nonNestables=array())
+    public function line($text, $nonNestables = array())
     {
         return $this->elements($this->lineElements($text, $nonNestables));
     }
@@ -1156,9 +1156,9 @@ class Parsedown
 
                 # cause the new element to 'inherit' our non nestables
 
-                foreach ($nonNestables as $non_nestable)
+                foreach ($nonNestables as $nonNestable)
                 {
-                    $Inline['element']['nonNestables'][] = $non_nestable;
+                    $Inline['element']['nonNestables'][] = $nonNestable;
                 }
 
                 # the text that comes before the inline
@@ -1192,8 +1192,10 @@ class Parsedown
 
         foreach ($Elements as &$Element)
         {
-            $Element['autobreak'] = isset($Element['autobreak'])
-                ? $Element['autobreak'] : false;
+            if ( ! isset($Element['autobreak']))
+            {
+                $Element['autobreak'] = false;
+            }
         }
 
         return $Elements;
@@ -1763,8 +1765,8 @@ class Parsedown
                 continue;
             }
 
-            $autoBreakNext = (isset($Element['autobreak']) && $Element['autobreak']
-                || ! isset($Element['autobreak']) && isset($Element['name'])
+            $autoBreakNext = (isset($Element['autobreak'])
+                ? $Element['autobreak'] : isset($Element['name'])
             );
             // (autobreak === false) covers both sides of an element
             $autoBreak = !$autoBreak ? $autoBreak : $autoBreakNext;
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -442,7 +442,7 @@ class Parsedown
 
     protected function blockFencedCode($Line)
     {
-        if (preg_match('/^(['.$Line['text'][0].']{3,})[ ]*([^`]+)?[ ]*$/', $Line['text'], $matches))
+        if (preg_match('/^(['.$Line['text'][0].']{3,}+)[ ]*+([^`]++)?+[ ]*+$/', $Line['text'], $matches))
         {
             $Element = array(
                 'name' => 'code',
@@ -486,7 +486,7 @@ class Parsedown
         }
 
         if (
-            preg_match('/^(['.preg_quote($Block['char']).']{3,})[ ]*$/', $Line['text'], $matches)
+            preg_match('/^(['.preg_quote($Block['char']).']{3,}+)[ ]*+$/', $Line['text'], $matches)
             and mb_strlen($matches[1]) >= $Block['openerLength']
         ) {
             $Block['element']['element']['text'] = substr($Block['element']['element']['text'], 1);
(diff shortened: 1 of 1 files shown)


Parsedown.php
```diff
@@ -182,29 +182,20 @@ protected function linesElements(array $lines)
...
-                    $shortage = 4 - mb_strlen($line, 'utf-8') % 4;
-
-                    $line .= str_repeat(' ', $shortage);
-                    $line .= $part;
-                }
+            for (
+                $beforeTab = strstr($line, "\t", true);
+                $beforeTab !== false;
+                $beforeTab = strstr($line, "\t", true)
+            ) {
```
PhrozenByte:
```php
while (($beforeTab = strstr($line, "\t", true)) !== false) {
    // ...
}
```

Parsedown.php
```diff
@@ -216,7 +207,7 @@ protected function linesElements(array $lines)
 
             if (isset($CurrentBlock['continuable']))
             {
-                $Block = $this->{'block'.$CurrentBlock['type'].'Continue'}($Line, $CurrentBlock);
+                $Block = $this->{"block{$CurrentBlock['type']}Continue"}($Line, $CurrentBlock);
```
PhrozenByte: readability:
```php
$methodName = 'block' . $CurrentBlock['type'] . 'Continue';
$Block = $this->$methodName($Line, $CurrentBlock);
```

aidantwoods: 👍 

As mentioned in:
> Adjustments to string concatenation syntax appear to offer no observable difference in Parsedown in my testing (despite historic benchmarks from the PHP community apparently indicating a difference).

I think I'll probably change this back to single quotes (as you've mentioned for readability), and since it doesn't offer any observable performance advantage in my testing. (I recall a long time ago that using `.` over interpolation had an observable negative difference – I guess things have changed now.)

Parsedown.php
aidantwoods: Thanks :D

Parsedown.php
PhrozenByte: I don't really aim for string interpolation vs. string concatenation here, but rather for defining a complex method name separately.

I don't think that there's a "real" performance difference between string interpolation and string concatenation. I personally prefer string concatenation because I think it's way more readable, especially for complex strings. However, others might say the complete different, so this is basically just a preference.

Parsedown.php
```diff
@@ -1156,9 +1138,9 @@ protected function lineElements($text, $nonNestables = array())
 
                 # cause the new element to 'inherit' our non nestables
 
-                foreach ($nonNestables as $non_nestable)
+                foreach ($nonNestables as $nonNestable)
                 {
-                    $Inline['element']['nonNestables'][] = $non_nestable;
+                    $Inline['element']['nonNestables'][] = $nonNestable;
                 }
```
PhrozenByte:
```php
$Inline['element']['nonNestables'] = array_merge($Inline['element']['nonNestables'], $nonNestable);
```

aidantwoods: Appending to undefined key creates it so we also have to check that the key exists with this method:
```php
$Inline['element']['nonNestables'] = isset($Inline['element']['nonNestables'])
    ? array_merge($Inline['element']['nonNestables'], $nonNestables)
    : $nonNestables
;
```

Still, I wouldn't be surprised if the behaviour currently in use gets warnings added to it at some point, and even if it doesn't, adding the check is still a good thing

Parsedown.php
```diff
@@ -1210,33 +1195,17 @@ protected function inlineText($text)
...
-            );
-        }
+        $safeText = self::escape($text, true);
+
+        $Inline['element']['rawHtml'] = preg_replace(
+            $this->breaksEnabled ? '/[ ]*+\n/' : '/(?:[ ]*+\\\\|[ ]{2,}+)\n/',
+            "<br />\n",
+            $safeText
+        );
+        $Inline['element']['allowRawHtmlInSafeMode'] = true;
```
aidantwoods: I'm still in two minds about this change – it's quite a bit faster because plaintext is pretty common and it saves us creating several arrays for each chunk of it, however it blinds any AST post-processor to the content structure. It also perhaps risks encouraging use of `rawHtml` + `allowRawHtmlInSafeMode` over the AST, which is less than ideal where it can be trivially avoided.

PhrozenByte: I'm not so sure whether breaks are *structural* elements or rather *styling* elements... Paragraphs are structural elements without any doubt, but single line breaks? Furthermore, isn't this a great example of AST processing adding unnecessary complexity? I don't think that we encourage people to use `rawHtml` + `allowRawHtmlInSafeMode` - it's rather a great example of how and under which circumstances it can and should be used. We shouldn't condemn `allowRawHtmlInSafeMode` :wink:

aidantwoods: > I'm not so sure whether breaks are *structural* elements or rather *styling* elements... 

Yes, you're correct in that they're used for style. I was more going for "structure" here to mean that breaks represent a different kind of object in the syntax tree than plaintext does. I don't really have much concern for *why* the elements are used (rather that they are elements) :p

I definitely don't want to condemn `allowRawHtmlInSafeMode`, but if someone needs to create something simple like `<p>foobar</p>`, then I'd much rather they wrote it like:

```php
$Element = array('name' => 'p', 'text' => 'foobar');
```

than

```php
$Element = array('rawHtml' => '<p>foobar</p>', 'allowRawHtmlInSafeMode' => true);
```

Especially so when `foobar` becomes a variable.

Even if someone does all the escaping correctly, it's much easier to not to have to if it isn't required.

Perhaps an example of how to use raw HTML in a safe way like this is valuable though.

PhrozenByte: > I definitely don't want to condemn allowRawHtmlInSafeMode, but if someone needs to create something simple like `<p>foobar</p>`, then I'd much rather …

> Perhaps an example of how to use raw HTML in a safe way like this is valuable though.

Full ACK :+1:

Issue #613, opened 2018-04-10
extending inline image causing all server memory to be consumed


runrevben:
```
private $baseImagePath = 'http://cdn.example.com/';

    protected function inlineImage($excerpt)
    {
        $image = parent::inlineImage($excerpt);

        $image['element']['attributes']['src'] = $this->baseImagePath . $image['element']['attributes']['src'];

        return $image;
    }
```

This is the example code from the site. I've created a class extending parsedown with the above code. It's consuming all php memory on the server until it throws a:

```Fatal error: Allowed memory size of 1073741824 bytes exhausted (tried to allocate 20480 bytes) in /srv/xxxxxx/includes/Parsedown.php on line 1828```

I upped the php memory limit to 1024mb as a test and it still consumed it. Any thoughts?

I've tried with parsdown v17 and the latest.

Seems to be tripping out in the pregReplaceElements function.

aidantwoods:
Is there a particular input you could give that causes this?

runrevben:
Thanks, I'm just looking now. I've been able to run one file through it successfully.. so let me run each file until I find the offending article. I'll post back shortly. Must be a specific URL that is causing issue.

runrevben:
OK, so here is the offending file:

https://raw.githubusercontent.com/livecode/livecode/9.0.0/docs/guides/HTML5%20Deployment.md

It doesn't have any images, but it does have an ! mark at the end of a line roughly half way down the document. Removing that ! allows the doc to parse correctly. As I understand it, markdown doesn't require special characters to be escaped. Additionally, it parses fine as long as I'm next extending ```inlineImage```.

aidantwoods:
This reminds me a little of https://github.com/erusev/parsedown/issues/546, if you check for `$image` being `null` and return early does this have any affect (as described in https://github.com/erusev/parsedown/issues/546#issuecomment-351102485)?

Otherwise I'll try debug this later today :)

runrevben:
Hey, that's perfect. That worked.. and it makes sense why it works too.

aidantwoods:
Awesome, feel free to update the Wiki with the bugfix too :) (I've just noticed that's where this example is coming from!) otherwise I'll fix it later when not on mobile.

runrevben:
Thanks. Added fix to https://github.com/erusev/parsedown/wiki/Tutorial:-Create-Extensions

aidantwoods:
Thanks! 🎉

Issue #546, opened 2017-12-12
Extends the data attributes for the image


staskamyshov:
I wrote a class to add a caption to the image.

```
class MyParsedown extends \Parsedown
  {
      protected function inlineImage($Excerpt)
      {
          $Inline = parent::inlineImage($Excerpt);
          $Inline['element']['attributes']['data-caption'] = $Inline['element']['attributes']['alt'];
          return $Inline;
      }
 }
```

As a result, it works as it should, but now the handler produces an error when processing the exclamation mark (for example, "Hello World!").

```
AH01215: PHP Notice:  Undefined index: name in /var/www/.../erusev/parsedown/Parsedown.php on line 1404
AH01215: PHP Notice:  Undefined index: extent in /var/www/.../erusev/parsedown/Parsedown.php on line 1047
AH01215: PHP Notice:  Undefined index: name in /var/www/.../erusev/parsedown/Parsedown.php on line 1404
AH01215: PHP Notice:  Undefined index: extent in /var/www/.../erusev/parsedown/Parsedown.php on line 1047
```
... and so on to infinity, and the Apache server does not respond.

What have I done wrong?

staskamyshov:
#

staskamyshov:
It's work! Thank @aidantwoods

aidantwoods:
Was there a reason this was reopened @staskamyshov?

PR #618, opened 2018-04-17
Drop eol and near eol versions of PHP

aidantwoods:
Per http://php.net/supported-versions.php PHP 7.0 is in security fix only mode and will cease to be supported in 7 months.

This PR is a proposal to start 2.0 from the latest actively supported PHP release, 7.1.

---

If merged, closes #619

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,15 @@
 language: php
 
-dist: trusty
 sudo: false
 
 matrix:
   include:
-    - php: 5.3
-      dist: precise
-    - php: 5.4
-    - php: 5.5
-    - php: 5.6
-    - php: 7.0
     - php: 7.1
     - php: 7.2
     - php: nightly
-    - php: hhvm
-    - php: hhvm-nightly
   fast_finish: true
   allow_failures:
     - php: nightly
-    - php: hhvm-nightly
 
 install:
   - composer install --prefer-dist --no-interaction --no-progress
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Better Markdown Parser in PHP
 * Super Fast
 * Extensible
 * [GitHub flavored](https://help.github.com/articles/github-flavored-markdown)
-* Tested in 5.3 to 7.2 and in HHVM
+* Tested in 7.1 to 7.2
 * [Markdown Extra extension](https://github.com/erusev/parsedown-extra)
 
 ### Installation
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         }
     ],
     "require": {
-        "php": ">=5.3.0",
+        "php": "^7.1",
         "ext-mbstring": "*"
     },
     "require-dev": {


aidantwoods:
See also: [On PHP Version Requirements](https://web.archive.org/web/20170809180405/https://blog.ircmaxell.com/2014/12/on-php-version-requirements.html).

aidantwoods:
I'll concede that drawing the line at 7.1 is pretty aggressive (hence why this is a PR and not a commit :) )

Do you have similar objections to instead drawing the line at 5.6? (is PHP 5.3 still maintained by anyone? What about 5.4, or 5.5?). I think dropping near EOL is aggressive, but dropping EOL is just sensible.

> In practice they simply won't upgrade to Parsedown 2.0 until they make the decision to release a new major version.

Bare in mind that this is *our* new major version. We don't get this decision again until we release a new major version. If Parsedown 1.* is anything to go by – we'll probably be supporting whatever versions we select here for a *long* time.

> If we raise Parsedown's requirements, we inevitably cut-off any depending project from Parsedown updates [...]

1. This is a major update, so we don't do this behind the scenes.
2. There is a section in the post I linked to that addresses whether we are abandoning users.

> IMO there should always be a strong technical reason to drop support for older versions

1. PHP 5.3 in particular is a real pain to support (see: [one of many silly PHP 5.3 specific workarounds](https://github.com/erusev/parsedown-extra/pull/117/commits/876ea91bfea0c38ac93c70034da7bf79d96bea06))
2. IMO we shouldn't give users a reason to keep using EOL software.

aidantwoods:
Neither this nor #619 need to be merged – these are created because I think this is a topic worth at least discussing :) (even if neither end up being merged)

PhrozenByte:
> Do you have similar objections to instead drawing the line at 5.6? (is PHP 5.3 still maintained by anyone? What about 5.4, or 5.5?). I think dropping near EOL is aggressive, but dropping EOL is just sensible.

I can't really draw a exact line, it's a consideration. For example, the latest RHEL 7 comes with PHP 5.4.16 (even though you can install a newer PHP version, they guarantee bug fixes for PHP 5.4 until 2024 and security fixes until 2027 - that's ridiculous, I know). RHEL 5 comes with PHP 5.1 - and is in it's "Extended Lifecycle Support" until 2020.

I don't say that we should support PHP 5.1. My point is that the "it's EOL" can't be a valid reason for dropping support, because that's not the full truth. IMO the only valid reason for dropping support of older versions are technical reasons.

> This is a major update, so we don't do this behind the scenes.

Being a major release (that is allowed to break BC), doesn't mean that it breaks BC for everyone. For example, if you don't use extensions (or just extensions that aren't affected by the latest changes, like Parsedown Extra) and don't have rather "special" contents, virtually nothing changes. The CommonMark specs were written with existing contents and parsers in mind.

> There is a section in the post I linked to that addresses whether we are abandoning users.
> IMO we shouldn't give users a reason to keep using EOL software.

It's the same as with #561: Parsedown is a library, we should leave this decision up to the application developer - as far as practicable, at least. That's the major difference between Parsedown being a library and end-user applications: We indeed have users that (hopefully) know what is going on and can make a informed choice on their own. It's about flexibility, not about encouriging users to do stupid things (like using a PHP version that is neither supported by PHP, nor their Linux distribution).

> PHP 5.3 in particular is a real pain to support (see: [one of many silly PHP 5.3 specific workarounds](https://github.com/erusev/parsedown-extra/pull/117/commits/876ea91bfea0c38ac93c70034da7bf79d96bea06))

This might be one of those technical reasons. There was a major step between PHP 5.3 and 5.4, it might be reasonable to draw the line there. Maybe we can get some more opinions about that?

aidantwoods:
> IMO the only valid reason for dropping support of older versions are technical reasons.

Here are a few :)
* Return types in PHP 7: up until this point there are zero type system guarantees about what objects you will receive when calling a function
* Scalar type hints in PHP 7: up until this point there is no way to assert the scalar types that we want to accept in a function unless we manually check them in the function body. Not being able to assert that you only want `string` as a variable is extremely annoying.
* `strict_types=1` in PHP 7: PHPs implicit type casts is mostly a headache which hides bugs in code
* Type hint collections with variadics using the `...` operator from PHP 5.6: this allows us to formally accept ordered collections of objects in a type safe way

These are all especially useful if we want to define interfaces

PhrozenByte:
These aren't technical reasons, but just a list of new features. Its use must either be required to implement something, or the alternative solution must significantly increase code's complexity, readability and/or costs, to be a technical reason.

aidantwoods:
Strictly speaking we *should*  be manually checking the return type of every function call and check the argument types being accepted in the body of every function – otherwise we're assuming types without guarentees, and this can go wrong.

For example, the Parsedown wiki has apparently had a type related bug in an example extension which causes a loop to never end under certain circumstances due to PHPs implicit type casting.

See: https://github.com/erusev/parsedown/issues/613, and https://github.com/erusev/parsedown/issues/546#issuecomment-351102485

Had types been validated, this error would have presented itself as a type error and not a hard to find infinite while loop.

PhrozenByte:
*Might* be true for the public API, but since we're planning to remove the support of extensions extending the Parsedown class and rather implementing a API anyway, this won't be necessary anymore (by the way: I was about to suggest using a class for AST elements anyway, arrays require too many `isset()` checks).

Don't get me wrong, I'm not opposing the usefulness of these features, they just aren't really *required*. Again, it's a matter of *consideration*. Don't take it to the extreme, common sense applies. If the interpretation (of what a "technical reason" is) is too strict, we end up supporting everything since PHP 4.0. If the interpretation is to loose, we end up supporting just today's nightly build of PHP.

aidantwoods:
> (by the way: I was about to suggest using a class for AST elements anyway, arrays require too many `isset()` checks).

Indeed 😉 
screen shot 2018-04-17 at 19 08 59

> Don't get me wrong, I'm not opposing the usefulness of these features, they just aren't really required. Again, it's a matter of consideration. Don't take it to the extreme, common sense applies. If the interpretation (of what a "technical reason" is) is too strict, we end up supporting everything since PHP 4.0. If the interpretation is to loose, we end up supporting just today's nightly build of PHP.

Yup 👍 I'm only making my case for why we might want these features – it's perfectly okay if we arrive at the conclusion that these aren't enough and don't outweigh cost of dropping support (just trying to put everything on the table as it were).

I think the case for types still has merit because even without extensions we will still be interacting with code for plugins (ideally we can make it really hard to get into situations where bugs are silently swept under the rug by PHP).
That said, we can implement [something like this](https://github.com/aidantwoods/SecureHeaders/blob/cc1421453de860fb5fb006a729d6b660eb110a09/src/Operations/AddHeader.php#L24) in worst case if we are supporting pre PHP 7, and declare the PHP docblock to define public API with respect to function signatures.

aidantwoods:
Some stats from Packagist on PHP version usage as of November 2017:

https://seld.be/notes/php-versions-stats-2017-2-edition

Version | Usage  | Diff from May 2017 
------- | ------ | ------------------
PHP 7.1 | 36.63% | (+18.99)
PHP 7.0 | 30.76% | (-5.36)
PHP 5.6 | 23.28% | (-8.16)
PHP 5.5 | 6.11%  | (-4.5)
PHP 5.4 | 1.51%  | (-1.6)
PHP 5.3 | 0.76%  | (-0.22)

Obviously this isn't necessarily representative of the minimum supported version of packages.

aidantwoods:
cc @rhukster I assume you'd be for maintaining compatibility with PHP 5.5 and up? (I note that Grav requires `>=5.5.9`)

aidantwoods:
So **excluding** any data from laravel (which supports 7.1 and later and has a disproportionate number of users compared to other dependencies), just over 50% of our indirect users are using us via a project that supports PHP 5.6 or later. (If we include laravel that number is almost 90%)

It's also worth noting that PHP 5.6 was shipped in Ubuntu 16.04 LTS, and that the latest LTS iteration that would have shipped with 5.5 is 14.04 LTS – which will itself hit EOL within a year. PHP 5.5 was EOL'd by the PHP team almost 2 years ago.

I also note that Grav (mentioned earlier in the thread as having a 5.5 requirement) is planning to move to 5.6 in their next release (https://github.com/getgrav/grav/issues/1977).

Looking at the Packagist user specific data, it is also clear that very few people are still using PHP 5.5 and below, (and there is in-fact a major push to PHP 7.1). This indicates that it will not be hard for dependencies to upgrade because most of their users are likely already using 5.6 and above. Yes it will probably require a major release, but raising the PHP requirement is something they will have inevitably to do at some point, so we're not exactly hurting them by giving them a reason to upgrade. If they're not already thinking about it, maybe we'll be the reason they do.

On top of all this, by drawing the line at 5.6, we'd also gain the `...` operator which would let us accept collections of objects in a type safe manner – which we will have to do: e.g. a collection of AST `Element` objects. The alternative to using the language syntax is to manually verify elements in an array. Yes, we could write a compact (to call) helper function, but using a built-in language feature is so much more readable, and likely faster.

#### So to summarise, if we were to draw a line at 5.6:
* We directly benefit from new features
  * Additionally by dropping 5.3 we also avoid a host of problems and stupid workarounds (e.g. [1](https://github.com/aidanwoods/parsedown/pull/6), [2](https://github.com/erusev/parsedown-extra/pull/117/commits/876ea91bfea0c38ac93c70034da7bf79d96bea06), [3](https://github.com/erusev/parsedown/pull/614/commits/7f4318dbdb5eed781a259ace08b5d5f31ac65208), ... to name a few)
* Most users have already moved into this support range (see Packagist user data)
* Our largest dependencies already support 5.6 and onwards (see Packagist dependency data)
* The dependencies that support pre 5.6 can likely raise their requirement without losing many users (see Packagist user data), as mentioned: they will have to do this eventually anyway (and sooner rather than later).

#### What happens to dependencies that support pre PHP 5.6?
In my mind, they should release a new major version starting from a non EOL'd PHP version.
Not only are most of their users already using one (see Packagist data), but they gain access to new features, as well as being another reason that their users (or their hosting providers) should install safe versions of PHP: within a year Ubuntu looks like it will cease support of PHP 5.5, the PHP team did so almost two years ago. If users haven't got a plan for upgrading: they're running out of time (or may have already).

#### But I use `$os` and it still supports `$oldPHPVersion`:

There might be other OSs that support (and only support) PHP 5.5 and below for longer, but let's be realistic: without the PHP team supporting these, you're likely almost exclusively getting any security updates from backports. It is unlikely anyone willing to report bugs will be looking for ones specific to PHP 5.5 and below. Much less will they know who to report these bugs to: they can't really report them to teams that don't support that version. Are they going to report them to all the minor OSs supporting only these old versions individually?

For users on OSs that offer later versions: guess what? To get new features you need to install updates (update your PHP installation – you'll not only be able to use newer software on it, but it'll be much safer too). For users on OSs that don't: they should update – or change – their OS.

PhrozenByte:
You're virtually arguing on the basis of usage statistics, the only problem is that the statistics you're using aren't relevant at all. It's like arguing that Windows is installed on roughly 90% of all computers, thus we don't have to support Linux :wink: 

The biggest problem is that Packagist is extremely rarely used to distribute software to end-users. I'm not sure why Grav is getting the "paradigm project", but just take Grav as an example: the usual way to install Grav is by using their pre-built package. These installations won't show up on Packagist. You can easily verify that by looking at their install count on Packagist: Do you really think Grav was installed just 22k times?

This leads us to another problem: Even the people that do use `composer`, usually don't use it on their productive server. They rather run `composer` it in their dev environment. Thus you're not getting statistics about the environments Parsedown is actually used on, but about dev environments.

The third problem is that you assume that all projects that depend on Parsedown use `composer`. Given that there are several Issues here on GitHub dealing with people downloading just the `Parsedown.php`, this assumption is insupportable.

> In my mind, they should release a new major version starting from a non EOL'd PHP version.

By assuming that Parsedown might be the reason for application developers to release a new major version of their software and raising their requirements because of Parsedown is pretty... daring :laughing: Since Parsedown isn't really security-relevant (not least because Parsedown wasn't able to deal with untrusted content until just a few weeks ago) they simply won't update Parsedown. This is bad, but this is reality - the bad reality.

> But I use `$os` and it still supports `$oldPHPVersion`:

So, you think how Ubuntu deals with PHP updates represents how enterprises like Red Hat deal with them? Red Hat plays in another league. RHEL and CentOS are one of the, if not the most wide-spread server distributions. We're not talking about "minor OSs" here. :confused: 

> Additionally by dropping 5.3 we also avoid a host of problems and stupid workarounds

As I've mentioned already there might be good technical reasons to drop support of PHP 5.3. However, there unfortunately wasn't a single persuasive argumen why PHP 5.4 and 5.5 should be dropped.

aidantwoods:
> You're virtually arguing on the basis of usage statistics[...]It's like arguing that Windows is installed on roughly 90% of all computers, thus we don't have to support Linux 😉

I'm not just arguing on the most popular PHP version, but the minimum supported version statistics of *our* dependencies. Furthermore I am arguing based on the presence of new features that will be beneficial for us to use. I make a point about metrics to discuss likely impact.

Also, *literally one of your first points* was about cutting off too many users/projects. Like, what?

On the basis of new features alone I'd love to just drop everything below 7.1, the usage metrics are helpful to see why we should support versions prior to that.

>[...] the only problem is that the statistics you're using aren't relevant at all.[...]

Really? Our dependencies are not at all relevant?

> I'm not sure why Grav is getting the "paradigm project"

I only mentioned this since I referred to it earlier in the thread as requiring 5.5 (when in-fact it is moving to requiring 5.6... and then to 7.1 in their next update for what it's worth – but I digress).

> The biggest problem is that Packagist is extremely rarely used to distribute software to end-users. [...]
> The third problem is that you assume that all projects that depend on Parsedown use `composer` [...]

Of course this is just a sample. When did I say I assume absolutely everyone uses composer & packagist? Unfortunately unless you can provide additional metrics on direct installations, discussing them is just a game of "what if's?" and guessing. I would *expect* the distributions of PHP versions to be similar – but who knows.
Maybe in 10 years we still won't know the answer, but we better support PHP 5.3 just incase projects are still using it that we don't know about, right? Why don't we just never move forward at all. Let's just do bugfixes so we don't need the new features from new PHP versions to make the codebase better... does this sound like a good idea?

> By assuming that Parsedown might be the reason for application developers to release a new major version of their software and raising their requirements because of Parsedown is pretty... daring 😆

Updating their requirements is something they have to do eventually. I doubt that Parsedown will cause them to do it. This is why I mention EOL and security. Most projects *do* (or *should*) care about security, by extension they should care if their users are running EOL software.
That said: the majority of projects by user mass **in the sample** will not have to update their requirements (this is the point in the metrics: to get an idea of impact).

> So, you think how Ubuntu deals with PHP updates represents how enterprises like Red Hat deal with them? Red Hat plays in another league. RHEL and CentOS are one of the, if not the most wide-spread server distributions. We're not talking about "minor OSs" here. 😕

No. The comment about minor OSs was specifically addressing the case of OSs that support **and only support** PHP 5.5 and below. If an OS supports a later PHP version, then it is a minor upgrade away and the user *should* upgrade.

It doesn't matter what "league" they play in: core PHP development has moved on. 

---

TL;DR: I think Parsedown will be better when using new PHP features. Most people *appear **in a sample*** to be using compatible versions of PHP. Most of **our** dependencies *appear **in a sample*** not to have to do anything to their min PHP version anyway. For the ones that do, they're likely already moving in that direction because 5.6 (a minor update from 5.x) was released almost 4 years ago.

Other than we leave a few people who can't update right now behind, is there any argument to support pre 5.6?

aidantwoods:
> there unfortunately wasn't a single persuasive argumen why PHP 5.4 and 5.5 should be dropped yet.

PHP 5.6 has type hinted variadics that facilitate type safe collections. Why should we keep them?

PhrozenByte:
> Why don't we just never move forward at all.

You're not getting the point. The point is that you can't use usage statistics to reason dropping support of platforms. Never, no matter what metrics you use. There is only *one* valid reason to drop support: Technical reason. There are indeed technical reasons to drop support of PHP 5.3, but none for PHP 5.4 and 5.5.

> Of course this is just a sample. When did I say I assume absolutely everyone uses composer & packagist? Unfortunately unless you can provide additional metrics on direct installations, discussing them is just a game of "what if's?" and guessing.

It's neither a *random* nor a (at least assumed) *representative* sample - and that's the reason why it isn't relevant *at all*. We know that this sample is highly skewed. This is nothing up to guessing, we *know* that, as described above (see https://github.com/erusev/parsedown/pull/618#issuecomment-382911466). It's exactly like arguing that Parsedown shouldn't support Linux because Windows is installed on 90% of all computers. The problem is not that the metrics are wrong, the problem is that you can't derive your assumptions from the given metrics.

You won't ever get any valid statement from this data about the *population*, this is statistically impossible. You can only make statements about the sample, but not about the population - because it isn't representative for the population. The only thing you can deduce from the given data is that there are at least 111 projects still supporting PHP 5.3, 5.4 and 5.5. You can deduce that this is more than half of all projects on Packagist (not of *all* projects using Parsedown!). You can deduce that there were hundreds of thousands (you unfortunately didn't provide the total number of downloads) of downloads on platforms still using PHP 5.3, 5.4 and 5.5. You can make statements about *absolute numbers*. You can't make statements about usage or usage shares, because this would require you to apply these metrics to the population. You can't deduce anything about the population. Again, this is statistically impossible because we *know* that the sample is highly skewed.

I'm just trying to tell you that you can't use the given information to make the statements you've made. This is completely independent of whether these statements are relevant for this consideration or not. They are not. Even though the absolute numbers rather support my statement (having a negative impact on more than 100 projects is bad), it is nevertheless irrelevant. Again, the only valid reason to drop support is a technical reason.

> The comment about minor OSs was specifically addressing the case of OSs that support and only support PHP 5.5 and below.

Just because a newer version is available doesn't mean that you can (or even should) use a newer version. This might be true for small private websites, but can't be further away from truth for enterprises. Never touch a running system unless you are *required to*. And there is no reason to do so until you have a **technical reason** - like that your Linux distribution ends support of your PHP version. However, since enterprise Linux distributions support PHP versions for ages this usually isn't going to happen anytime soon. The most common reason is that enterprises replace old servers with newer ones, using newer software. But this happens just bit by bit.

> PHP 5.6 has type hinted variadics that facilitate type safe collections.

See https://github.com/erusev/parsedown/pull/618#issuecomment-382088489

rhukster:
Sorry to be so late in replying re. Grav, but we are actualy moving quite rapidly ourselves to drop EOL PHP releases: https://www.getgrav.org/blog/raising-php-requirements-2018

* Grav 1.5 (PHP 5.6.4+) we plan on releasing in a month or two
* Grav 1.6 (PHP 7.1.3+)should be released before the end of the year.

Issue #306, opened 2015-04-02
Giving a header an ID


IngwiePhoenix:
Hey.

Will Parsedown ever support

```
# Heading {#some-id}
```

or similar?

Kind regards, Ingwie.

hkdobrev:
@IngwiePhoenix This is a feature from [Markdown Extra](https://michelf.ca/projects/php-markdown/extra/#spe-attr) which [Parsedown Extra](https://github.com/erusev/parsedown-extra) supports.

IngwiePhoenix:
Is it the same API? Or do I need to change anything in my code to use Parsedown Extra?

hkdobrev:
@IngwiePhoenix It's the same. It's just a simple extension of the Parsedown class. You just need to include the `ParsedownExtra.php` file (or if you are using Composer, just `composer require erusev/parsedown-extra`) and make an instance of the `ParsedownExtra` class.

IngwiePhoenix:
Awesome. I just changed to ParsedownExtra on my project and its awesome. :) Thanks!

aidantwoods:
Closing since the extension fills this role

sbrl:
Is it possible to add ids automagically with ParsedownExtra?

hellocosmin:
@sbrl What would the automagically-added ID look like? Consider this heading:

```markdown
# I'm not (exactly) sure how to create/build my own ID
```

Yeah, it could manipulate the string into some sort of slug, but how could ParsedownExtra know what kind of 'slug' you want, or how long the text of the ID should be?

For example, maybe you want it to output:

```html 
<h1 id="im-not-exactly-sure-how-to-create-build-my-own-id">I'm not (exactly) sure how to create/build my own ID</h1>
```

OK. Then again, maybe someone would prefer underscores:

```html 
<h1 id="im_not_exactly_sure_how_to_create_build_my_own_id">I'm not (exactly) sure how to create/build my own ID</h1>
```

Surely, for long text, everyone would like short IDs, maybe something like:

```html 
<h1 id="create-build-own-id">I'm not (exactly) sure how to create/build my own ID</h1>
```

That's why this belongs to userland. Either create an extension that 'slugifies' the text as you need it, or just use ParsedownExtra as intended:

```markdown
# I'm not (exactly) sure how to create/build my own ID {#custom-id}
```

sharpicx:
[...]

PR #620, opened 2018-04-19
Implemented Header ID tags

detain:


aidantwoods:
If you're interested in adding this I think an extension is probably best suited to this? It's maybe not entirely obvious how to come up with a method of generating the ID that'll suit everyone (some discussion: https://github.com/erusev/parsedown/issues/306#issuecomment-377683857).
Some might prefer methods which honour the original title as best as possible and don't care about typing it (e.g. some encoding with URL safe base64), others might prefer to strip out non-letter characters, or to replace them with `-`s, etc...

PR #621, opened 2018-04-23
Tilde characters may be escaped

paukenba:
According to http://spec.commonmark.org/0.28/#example-289 and https://github.github.com/gfm/#example-299 one should be able to escape a list of punctuation characters. Tilde is one of them


aidantwoods:
Thanks! :)

PR #624, opened 2018-04-28
Fix function inlineCode trim spaces

workschan:
Just trim the end of line spaces, otherwise the start of next line will has a space.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1258,7 +1258,7 @@ class Parsedown
         if (preg_match('/^('.$marker.'+)[ ]*(.+?)[ ]*(?<!'.$marker.')\1(?!'.$marker.')/s', $Excerpt['text'], $matches))
         {
             $text = $matches[2];
-            $text = preg_replace("/[ ]*\n/", '', $text);
+            $text = preg_replace("/[ ]*\r\n/", ' ', $text);
 
             return array(
                 'extent' => strlen($matches[0]),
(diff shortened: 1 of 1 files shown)


aidantwoods:
Wouldn't this change something like
```markdown
`foo
bar`
```

into parsing like

```html
<code>foobar</code>
```

instead of

```html
<code>foo bar</code>
```

I don't think this is correct (per spec)? ([reference parser](https://spec.commonmark.org/dingus/?text=%60foo%0Abar%60))

workschan:
I test it, it keep the newline
<code>foo
bar</code>
not you say
<code>foobar</code>

aidantwoods:
> it keep the newline

I just tested it locally, and it seems to remove it?

screen shot 2018-05-05 at 14 07 02

aidantwoods:
Hmm.. Parsedown should be converting `\r\n`s and `\r`s to `\n` when it receives input to avoid having to consider different line endings in latter code. Are you by any chance using the `->line` method in your testing (as opposed to `->text`)? (I'm just noticing that this would skip the line ending standardisation)

aidantwoods:
Instead of the changes to the regex, if you were to add the following
https://github.com/erusev/parsedown/blob/dfab7240a4081e4bdc818792803526840ab63542/Parsedown.php#L42-L43

into the `line` method, I think this should ensure that different line endings are treated the same (and should prevent any discrepancies in output caused by differing line ending sequences).

PR #644, opened 2018-06-11
Allow setting target for URLs

jukra:
I needed to have URLs converted to links with target="_blank" specified. This adds such functionality, if ok, I could update the docs as well for this?


kminchev:
Seconded.

PR #645, opened 2018-06-11
Updated to PSR-1 & PSR-2

BenjaminHoegh:
Updated to follow the PSR-1, PSR-2 rules

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1,6 +1,4 @@
 <?php
-
-#
 #
 # Parsedown
 # http://parsedown.org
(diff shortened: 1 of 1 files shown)


BenjaminHoegh:
lol, #TravisCI it works you monkeys

aidantwoods:
I think we could incorporate some of these (e.g. adding the explicit `public` modifier to public functions), although I don't think most of these changes would work for right now:

Unfortunately a lot of these are breaking changes (e.g. changing the names of instance variables and constants) – this is why the Travis build is failing.

On camelCasing local variables: in Parsedown the existing convention is to use PascalCase for arrays that have some expected structure (e.g. line and element structures). This could perhaps be dropped in a new major release when transitioning to objects to place type safety constraints on these structures, but as of now this notation provides readability value and helps communicate the writers intent.

IMO while Parsedown is in a single file, liberal use of whitespace also helps readability by sectioning off relevant disjoint components or sections of a function (I'm not sure why it is being removed even between function definitions).

On brace placement for `if`s and loops, IMO putting the brace on the next line helps clarify sectioning better (Parsedown's `if` clauses along with the body of both `if` and loop constructs are typically non-trivial), and so I think the need for clearer sectioning is justified.

BenjaminHoegh:
@aidantwoods what about namespace? should it be added?

aidantwoods:
For the next major version, likely yes :) See https://github.com/erusev/parsedown/pull/617

PR #668, opened 2018-09-19
remove redundant header level cap

itshoro:
As we already determine the heading level with strspn and return if it is greater than 6 there is no reason to use the min function here.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -541,7 +541,7 @@ class Parsedown
 
         $Block = array(
             'element' => array(
-                'name' => 'h' . min(6, $level),
+                'name' => 'h' . $level,
                 'handler' => array(
                     'function' => 'lineElements',
                     'argument' => $text,


aidantwoods:
Took a while to track down which of those was introduced first, and the [commit that introduced the second one](https://github.com/erusev/parsedown/commit/06135cd75a61e8c31429b8987a69b61f4a77ab34#diff-ed0d3da57330712681e64f838087ea47R488) :p

The best reason I can come up with for the `$level > 6` early return as well as the `min` both existing is that `min` was there first, but the `$level > 6` check was needed so that if the or branch in the following executes, then it can be assumed that `$level <= 6`: 

https://github.com/erusev/parsedown/blob/06135cd75a61e8c31429b8987a69b61f4a77ab34/Parsedown.php#L488

In any case—you're completely correct, `min` is redundant and can be removed :)

PR #670, opened 2018-10-02
Add "tel:" to whitelist

tillkruss:
Unless there is a way to customize the "safe links" whitelist, we should allow `tel:` links.


aidantwoods:
It's customisable at the extension level, but this seems generic enough to add in to the bundled whitelist regardless of any ability to customise anyway :)

PR #677, opened 2018-11-10
Added support for SyntaxHighlighter

miguel10rc:
transform ```xxx to <pre class="brush:xxx">
`<pre class="brush:xxx">`

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -470,7 +470,12 @@ class Parsedown
 
         if ($infostring !== '')
         {
-            $Element['attributes'] = array('class' => "language-$infostring");
+            $attrb = array('class' => 'brush:'.$infostring);
+            $Element = null;
+        }
+        else
+        {
+            $attrb = null;
         }
 
         $Block = array(
@@ -478,6 +483,7 @@ class Parsedown
             'openerLength' => $openerLength,
             'element' => array(
                 'name' => 'pre',
+                'attributes' => $attrb,
                 'element' => $Element,
             ),
         );
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -119,6 +119,15 @@ class Parsedown
         'steam:',
     );
 
+    function setSyntaxHighlighter($syntaxHighlighter)
+    {
+        $this->syntaxHighlighter = $syntaxHighlighter;
+
+        return $this;
+    }
+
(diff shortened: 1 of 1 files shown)


brainexcerpts:
Nice PR! Unfortunately doesn't play too well with the latest version of Parsedown.php.
I tried to fix the PR but Parsedown.php now encloses code blocks with an extra <code></code> tag and although adding the class to the parent <pre></pre> was easy in Parsedown.php I could not figure out how to avoid the insertion of the code tags. 

Instead I propose to patch shCore.js of syntax `SyntaxHighlighter`:
 
```js

        highlight : function(globalParams, element)
	{
		function toArray(source)
		{
			var result = [];

			for (var i = 0; i < source.length; i++)
				result.push(source[i]);

			return result;
		};

		var elements = element ? [element] : toArray(document.getElementsByTagName(sh.config.tagName)),
			propertyName = 'innerHTML',
			highlighter = null,
			conf = sh.config
			;
                 // BEGIN PATCH
		// remove code tags in elements if present:
		// <pre><code>...</code></pre> -> <pre>...</pre>
		for (var i = 0; i < elements.length; i++)
		{
			var element = elements[i];
			if (element.tagName != undefined && element.tagName.toLowerCase() == 'pre' && element.childNodes.length == 1)
			{
				var child = element.firstChild;

				if (child.tagName != undefined && child.tagName.toLowerCase() == 'code' && child.childNodes.length == 1)
				{
					// place content of the code tag into the pre tag and remove the code tag
					const code = child.innerHTML;
					// convert to html entities
					let childClass = child.className;
					// extract language from class "language-xxx"
					const lang = childClass.match(/language-(\w+)/);
					element.innerHTML = code;
					if (lang) {
						element.className = ("brush:" + lang[1]);
					}
				}
			}
			elements[i] = element;
		}
                //END PATCH
		// support for <SCRIPT TYPE="syntaxhighlighter" /> feature
		if (conf.useScriptTags)
			elements = elements.concat(sh.utils.getSyntaxHighlighterScriptTags());
```

This will delete the code tags and read the language class and put it in the pre tag. So far this solution seems to work well.

Since Parsedown.php is not about syntax highlighting and seeing that it now provides proper class names for code blocks to allow further post processing and highlighting, I don't think `Parsedown.php` is that right place to add such feature (at least in the official project).

PR #678, opened 2018-11-12
Added setLinksOpenNewTab

WilliamVenner:
A simple variable that, when `true`, sets `target` attribute to `_blank` when generating an `a` tag for markdown links. Great utility for anyone who wants to enforce all links to open in a new tab without having to create an extension.


kminchev:
:thumbsup:

PS: why you capitalize the namespace, eg. Erusev\Parsedown instead of erusev/parsedown, like the repo name?

PR #715, opened 2019-05-18
Added a hook to change blocks on-the-fly

scorninpc:
I see alot issue about change text and block style, like a Geshi or new styles

I see it's confuse create Extras for every block we need to change. I'm big fan of hooks, so, I implemented something like that to do any change on-the-fly.

Something like that:

```php
$Parsedown = new Parsedown();

// Hook for change element blocks
$Parsedown->setTextHook(function($Element) {

	// Look for code block
	if($Element['name'] === "code") {
		$geshi = new Geshi($Element['text'], str_replace("language-", "", $Element['attributes']['class']), "/media/backup/www/andor.com.br/library/General/Library/Geshi/geshi/");
		$geshi->set_header_type(GESHI_HEADER_NONE);
		unset($Element['text']);

		$Element['rawHtml'] = $geshi->parse_code();
		$Element['allowRawHtmlInSafeMode'] = TRUE;
	}

	// ANOTHER ONE TO CHANGE
	elseif (1) {

	}

	// Return the same element block
	return $Element;
});
```

I don't know if you like or it's can be useful for you. But, was already coded!

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -13,11 +13,14 @@
 #
 #
 
-class Parsedown
+class Parsedown 
 {
     # ~
 
     const version = '1.8.0-beta-7';
+
+    # ~ 
+    protected $texthook = NULL;
 
     # ~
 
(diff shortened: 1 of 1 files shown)


onelikeandidie:
This sounds like a great feature to have, I also wanted to hook into Parsedown without modifying it directly to add some extra features. Shame this was request was closed!

PR #731, opened 2019-10-24
add class option

Linnzh:
I added explicit class name addition options in some places.


Linnzh:
Is this library not allowed to add class names to HTML? In other words, can't modify the original label?

Linnzh:
Why not explicitly provide an opportunity to add a class name? Do you want to put the generated HTML inside a <div> tag with a controllable class name?

Issue #732, opened 2019-10-24
Add support for PHP 7.1+ and strict types


janbarasek:
Hello,

I suggest add support for all new PHP 7 features like strict types, input and return types, use new array syntax and more.

This upgrade can be released as new package master version and used in `composer.json`.

I think keep progress with current PHP is very important for new projects.

If you want upgrade this package, please let me know and I will send pull request.

Thank you and have a lovely day. :)

aidantwoods:
See #708. 

I'd actually really appreciate a review of that PR, to avoid just making changes no one really wants or agrees with.

janbarasek:
@aidantwoods Ok, I will check it. I think you can close this issue and we will continue in #708.

Thanks.

PR #733, opened 2019-10-24
Support for PHP 7

janbarasek:
Solve #732

- Added support for PHP 7
- Used all new PHP 7 features
- Strict types
- `declare(strict_types=1);`
- Input and return data types in methods
- Whole code refactoring and fix better readable
- Refactoring to tab and PSR-4 convention
- Add public method visibility

This pull request and change is maximally back compatible, but I renamed all variable names and keep different property name convention.

Please let me know which changes are acceptable to you, because I think this package should keep progress.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1,1994 +1,1404 @@
 <?php
 
-#
-#
-# Parsedown
-# http://parsedown.org
-#
-# (c) Emanuil Rusev
-# http://erusev.com
-#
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
How much slower is this code than the original (on the same PHP version)?

janbarasek:
@GrahamCampbell Sorry for late answer, but I must make some performance testing.

Speed is same +/- 5 %, my suggestion does not change speed.

PR #740, opened 2019-12-01
make it possible to disable block and inline types

lsmith77:
this is just PoC


lsmith77:
would you be open to such a feature?

PR #742, opened 2019-12-28
Revert BC break

GrahamCampbell:
Allows the parsedown extra tests to run.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,4 @@
 # Ignore all tests for archive
-/test               export-ignore
 /.gitattributes     export-ignore
 /.gitignore         export-ignore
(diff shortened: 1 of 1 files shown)


aidantwoods:
I'm not sure this is a BC break, given that the PR I think you're referring to [explicitly swaps]( https://github.com/erusev/parsedown-extra/pull/141/files#diff-354f30a63fb0907d4ad57269548329e3R25) `--prefer-source` for `--prefer-dist`, which results in the test directory being absent on install: causing the test failure.

Downloading Parsedown's test suite on every install is likely a waste of everyone's time; for those that *really* want it there is `--prefer-source`.

Original PR adding this is here: https://github.com/erusev/parsedown/pull/406

PR #754, opened 2020-02-23
add a namespace for class Parsedown

parsampsh:
Add a namespace "Parsedown"

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -13,6 +13,8 @@
 #
 #
 
+namespace Parsedown;
+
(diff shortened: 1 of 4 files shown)


danilopolani:
Up here. Why the package removed the namespace? I saw in a previous old PR (2018) where a namespace was added for PSR-4, but now it has been removed again (?)

Bizarrus:
`+1`

Especially if you want to use it as an external library, a namespace is necessary.

I also wonder why there is no namespace assignment.

Issue #633, opened 2018-05-30
make internal links with this library


MahdiPishguy:
after installing erusev/parsedown library package i want to make internal links with `Markdowns` like with this code:

    $Parsedown = new Parsedown();
    echo $Parsedown->text("* [Hello](#hello)  
                                * [Hello World](#hello-world)
                                * [Another section](#new-section)
                                
                                ## Hello
                                ### Hello World
                                ## New section");

in this code i expect `## New section` must be have `id`, but after show this markdown this part of text doesn't have any `id` to make simple internal link

tovic:
Try `ParsedownExtra` extension then do this:

~~~ .markdown
### New Section {#new-section}
~~~

aidantwoods:
Headers being given IDs while useful in some contexts (e.g. a `*.md` file as rendered by GitHub), isn't desirable in others (e.g. a comment as rendered by GitHub). The default (by spec) is not to have these, but the option is there to have them via an extension as @tovic has mentioned :) (see also: [how would we make the ID a useful format for everyone?](https://github.com/erusev/parsedown/issues/306#issuecomment-377683857) – not having it by default is a decent way of avoiding having to make a choice on someones behalf).

PR #765, opened 2020-04-29
Create ID's for Header elements so they can be referenced in anchor tags

### LINKED ISSUE #710: Table of Contents doesn't work
For example, try the readme file here:

https://github.com/CodepadME/laravel-tricks

In GitHub, it works as expected, because:

`## Title`

is translated to:

<h2><a id="title" class="anchor" aria-hidden="true" href="#title">{svg removed for brevity}</a>Title</h2>

While this library translates it to:

<h2>Title</h2>

So, basically it's missing the anchor.

netniV:
This allows rendered markdown to be compatible when viewed via parsed markdown or GitHub

Closes #710

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -554,13 +554,17 @@ class Parsedown
 
         $text = trim($text, ' ');
         $link = strtolower(str_replace(' ','-',$text));
+	$attr = array();
+	if (!empty($link)) {
+		$attr = array(
+                    'id' => $link,
+                );
+	}
 
         $Block = array(
             'element' => array(
                 'name' => 'h' . $level,
-                'attributes' => array(
-                    'id' => $link,
-                ),
+                'attributes' => $attr,
                 'handler' => array(
                     'function' => 'lineElements',
                     'argument' => $text,
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -553,15 +553,19 @@ class Parsedown
         }
 
         $text = trim($text, ' ');
+        $link = strtolower(str_replace(' ','-',$text));
 
         $Block = array(
             'element' => array(
                 'name' => 'h' . $level,
+                'attributes' => array(
+                    'id' => $link,
+                ),
                 'handler' => array(
                     'function' => 'lineElements',
                     'argument' => $text,
                     'destination' => 'elements',
-                )
+                ),
             ),
         );
 
@@ -1992,3 +1996,4 @@ class Parsedown
                    'wbr', 'time',
     );
 }
+
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -553,7 +553,7 @@ class Parsedown
         }
 
         $text = trim($text, ' ');
-        $link = strtolower(str_replace(' ','-',$text));
+	$link = preg_replace('/[^\p{L}\p{N}\p{M}-]+/u', '', mb_strtolower(mb_ereg_replace(' ','-',$text)));
 	$attr = array();
 	if (!empty($link)) {
 		$attr = array(
--- a/test/ParsedownTest.php
+++ b/test/ParsedownTest.php
@@ -54,8 +54,10 @@ class ParsedownTest extends TestCase
         $this->Parsedown->setStrictMode(substr($test, 0, 6) === 'strict');
 
         $actualMarkup = $this->Parsedown->text($markdown);
+        $actualMarkup = str_replace("\r\n", "\n", $actualMarkup);
+        $actualMarkup = str_replace("\r", "\n", $actualMarkup);
 
-        $this->assertEquals($expectedMarkup, $actualMarkup);
+        $this->assertEquals(trim($expectedMarkup), trim($actualMarkup));
     }
 
     function testRawHtml()
--- a/test/data/atx_heading.html
+++ b/test/data/atx_heading.html
@@ -1,13 +1,13 @@
-<h1>h1</h1>
-<h2>h2</h2>
-<h3>h3</h3>
-<h4>h4</h4>
-<h5>h5</h5>
-<h6>h6</h6>
+<h1 id="h1">h1</h1>
+<h2 id="h2">h2</h2>
+<h3 id="h3">h3</h3>
+<h4 id="h4">h4</h4>
+<h5 id="h5">h5</h5>
+<h6 id="h6">h6</h6>
 <p>####### not a heading</p>
-<h1>closed h1</h1>
+<h1 id="closed-h1">closed h1</h1>
 <h1></h1>
 <h2></h2>
-<h1># of levels</h1>
-<h1># of levels #</h1>
-<h1>heading</h1>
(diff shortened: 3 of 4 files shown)


Parsedown.php
```diff
@@ -553,15 +553,19 @@ protected function blockHeader($Line)
         }
 
         $text = trim($text, ' ');
+        $link = strtolower(str_replace(' ','-',$text));
```
Ayesh, line 556: I think a more elaborate "sluggify" mechanism would be better here. For example, `strtolower(preg_replace('/[^A-z-_]+'/, '-', $text))` to replace all non A-z/-/_ characters with a dash.

netniV: Yeah, I'm not sure what GitHub's method of replacement is but think we should probably be using:

```php
strtolower(preg_replace('/[^A-Za-z0-9\-_]+'/, '-', $text))
```

Would you agree with that one?

netniV: https://regex101.com/r/jWgxEc/1 - This is an example but I also had to include \v to prevent new lines from being picked up in the example.

Ayesh: You are right, this approach is a better start. GitHuv flavor looks quite complicated and forgiving, so I put my regex hat. 

Test gist: https://gist.github.com/Ayesh/250786888e7f4f146117aa96afcd0071
Suggested regex: [`[^\p{L}\p{N}\p{M}-]+`](https://regex101.com/r/hDI9DM/1)
Implementation: https://3v4l.org/melBh

The third heading in the Gist contains letters from my Sinhalese language. It has letters in both `\p{L}` (letters) and `\p{M}` (marks). These two, combined with `\p{N}` (numbers) and the dash gives us the negating regex that strips out everything else. 

This library already requires mbstring extension, so I don't see a problem with using `mb_strtolower`.

netniV: OK, seems better 👍 I love it when a plan comes together.

netniV: I have just pushed a change, which takes parts of yours but changes str_replace as well since we are talking potential multi-byte characters.  Seems to work with the gist, let me know what you think.

Ayesh:
I was reading how `league/commonmark` does the slugs, and it turns out to be stripping a lot characters as well, so I sent a PR https://github.com/thephpleague/commonmark/pull/467

Parsedown.php
```diff
@@ -553,7 +553,7 @@ protected function blockHeader($Line)
         }
 
         $text = trim($text, ' ');
-        $link = strtolower(str_replace(' ','-',$text));
+	$link = preg_replace('/[^\p{L}\p{N}\p{M}-]+/u', '', mb_strtolower(mb_ereg_replace(' ','-',$text)));
```
Ayesh, line 556: Do we absolutely need the `mb_ereg_replace()` here? It's not a regular expression, so we should be fine using `str_replace(' ', '-', $text)`. GFM seems to convert consecutive spaces to consecutive dashes as well. 

I'm not a decision maker, but as someone who spends a lot of time processing text and with this library, this looks great to me otherwise. Thank you!

Ayesh (approved):

Parsedown.php
netniV: I think so, as if you consider a four byte character, one of those could match a space.  Thus it would invalidate that character or change it to something it's not meant to be.  I never deal with them but if we really want to make it universal, I would suggest we do.

netniV:
> I was reading how `php-league/commonmark` does the slugs, and it turns out to be stripping a lot characters as well, so I sent a PR [thephpleague/commonmark#467](https://github.com/thephpleague/commonmark/pull/467)

Didn't even know about that one 👍

netniV:
so the tests are failing now because the leading # is no longer applicable.  The question is, should trim the leading hyphen?

aidantwoods:
Hi @netniV and @Ayesh thanks for the great work you've done towards these changes to arrive at a very sensible slug function :)

The topic of slugs (or header ID's) have come up a few times in the past [(e.g.)](https://github.com/erusev/parsedown/issues/633#issuecomment-394092744), and the position has previously been that there [wasn't a universal choice that everyone using the library would want](https://github.com/erusev/parsedown/issues/306#issuecomment-377683857), so to avoid making the choice for everyone we'd leave it alone and up to extensions.

I've built upon the changes you've worked on here in PR #767, which modifies these changes for the new `2.0.x` branch, as well as adding the option for a user-defined slug function (so that anyone who doesn't like the default can specify their own).

@netniV I've listed you as a co-author in the commit to that PR (which I hope is okay?) because most of the work has been done here in arriving at a sensible slug function to use.

netniV:
Looks good. I added one comment as per our discussions here over str_replace. More involved is your change but I was thinking it should be an optional addition as it could affect other elements due to a duplication of ID

The only other thought is with trimming a trailing and leading hyphen as that seems wrong to keep those to me.

netniV:
Does this mean this PR should now be closed as it will only be applied to the 2.0 version?

cpeel:
@aidantwoods - is there a timeframe for when 2.x will be released? I'd rather have my users use the auto IDs for headers than enable ParsedownExtra.

netniV:
It would be nice if this PR could be merged into the 1.8.x branch and a 1.8.1 released.  Either that, or have 2.0 released.  At the moment, I can't get these changes through composer without publishing my own version of the package.

Ayesh:
I don't think drastic changes like this would be realistically merged to the 1.x branch. What I do is simply extend the `Parsedown` class, and override the necessary parts. It turned out a good approach long term because I wanted to add some extra markup on my own too.

netniV:
Let me see if I can manage that.  I did something for images so it may be possible.  

Personally, I wouldn't call it a drastic change, but I can see ID conflicts could break things so I see where you are coming from.  I would submit a forked version with it in, but that seems a waste for such a small feature.

netniV:
Firstly, I think it would be good to have this 1.8.0-beta7 as a full release given there has been no other changes in the past year that I can see (Aug 2020).  Requiring a beta version in a composer.json looks ... well bad to me.  The reason that I need to use the beta version is for handler array usage.

Now that I have fixed on the latest versions, I am still getting a problem which is due to changes in this beta version causing an issue with your parsedown-extra class.

https://github.com/netniV/ParsedownID/issues/1

netniV:
@Ayesh or @aidantwoods do you have the ability to sort out the current mess of Parsedown vs Parsedown-Extra ?  Due to the beta changes that haven't been published as a version, parsedown-extra relies on older versions.  But to use the newer array-based handler definitions, you need this beta version and parsedown-extra apparently hasn't been updated in two years so can't handle it.

aidantwoods:
It's been a while, but now very close to a finished product for 2.0 in Parsedown (https://github.com/erusev/parsedown/pull/708) and ParsedownExtra (https://github.com/erusev/parsedown-extra/pull/172). I'd appreciate any feedback you have :)

PR #771, opened 2020-06-26
Add tel: linktype to whitelist

TomWor:
Please consider adding this link type to the whitelist.
It's commonly recognized among modern browsers and especially convenient on mobile devices.


TomWor:
I just saw that this was added already recently in master. Nevermind.

PR #819, opened 2021-11-21
Add comma to specialCharacters list

grimalschi:
This solves wrong behaviour with escaped commas.

You can see this wrong behaiviour using "1\. Hello\, world!" phrase at https://parsedown.org/demo

Screenshot 2021-11-21 at 07 30 55

This is important because this kind of escape ("1\. Hello\, world!") produced by one of most famous markdown editors - TUI editor.

If you write list without marking its as a real list, you will get this kind of escaping and then if you use result in Parsedown, you see "\," instead of just commas.

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1961,7 +1961,7 @@ class Parsedown
     # Read-Only
 
     protected $specialCharacters = array(
-        '\\', '`', '*', '_', '{', '}', '[', ']', '(', ')', '>', '#', '+', '-', '.', '!', '|', '~'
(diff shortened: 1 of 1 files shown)


grimalschi:
CommonMark also "unescapes" commas:
https://spec.commonmark.org/dingus/?text=%5C*not%20emphasized*%0A%5C%3Cbr%2F%3E%20not%20a%20tag%0A%5C%5Bnot%20a%20link%5D(%2Ffoo)%0A%5C%60not%20code%60%0A1%5C.%20not%20a%20list%0A%5C*%20not%20a%20list%0A%5C%23%20not%20a%20heading%0A%5C%5Bfoo%5D%3A%20%2Furl%20%22not%20a%20reference%22%0A%5C%26ouml%3B%20not%20a%20character%20entity%0A

PR #821, opened 2021-11-28
Documentation 2.0

aidantwoods:
This isn't yet complete, but is a start at documenting some of the updated internals.

### DIFF
--- a/docs/Migrating-Extensions-v2.0.md
+++ b/docs/Migrating-Extensions-v2.0.md
@@ -44,6 +44,16 @@ $Parsedown = new Parsedown(ParsedownExtra::from(ParsedownMath::from(new State)))
 $Parsedown = new Parsedown(ParsedownMath::from(ParsedownExtra::from(new State)));
 ```
 
+In the above, the first object that we initialise the chain of composed extensions is the `State` object. This `State`
+object is passed from `ParsedownExtra` to `ParsedownMath`, and then finally, to `Parsedown`. At each stage new
+information is added to the `State`: adding or removing parsing instructions, and to enabling or disabling features.
+
+The `State` object both contains instructions for how to parse a document (e.g. new blocks and inlines), as well as
+information used throughout parsing (such as link reference definitions, or recursion depth). By writing `new State`,
+we create a `State` object that is setup with Parsedown's default behaviours, and by passing that object through
+different extensions (using the `::from` method), these extensions are free to alter, add to, or remove from that
+default behaviour.
+
 ## Introduction to the `State` Object
 Key to Parsedown's new composability for extensions is the `State` object.
 
--- a/docs/Migrating-Extensions-v2.0.md
+++ b/docs/Migrating-Extensions-v2.0.md
@@ -68,6 +68,8 @@ functionality from different authors as they like. For example, a feature of Par
 A user could import *only* the abbreviations feature from ParsedownExtra by using the following:
 
 ```php
+use Erusev\Parsedown\State;
+use Erusev\Parsedown\Parsedown;
 use Erusev\ParsedownExtra\Features\Abbreviations;
 
 $State = Abbreviations::from(new State);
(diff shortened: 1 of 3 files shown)


docs/Migrating-Extensions-v2.0.md
```diff
@@ -0,0 +1,210 @@
...
+
+Here, ParsedownExtra is *composed* with Parsedown, but does not extend it.
+
+A key feature of *composability* is the ability to compose *multiple* extensions together, for example another
+extension, say, `ParsedownMath` could be composed with `ParsedownExtra` in a user-defined order.
+
+This time using the `::from` method, rather than the convinence constructor provided by `ParsedownExtra`.
+
+```php
+$Parsedown = new Parsedown(ParsedownExtra::from(ParsedownMath::from(new State)));
```
BenjaminHoegh, line 40: Some user may gets confused by the `new State` as they there is no information in where this gets created in the first place.

docs/Migrating-Extensions-v2.0.md
```diff
@@ -0,0 +1,210 @@
...
+`StateBearer`. This will allow your users to cherry-pick specific pieces of functionality and combine it with other
+functionality from different authors as they like. For example, a feature of ParsedownExtra is the ability to define and expand
+"abbreviations". This feature is self-contained, and does not depend on other features (e.g. "footnotes").
+
+A user could import *only* the abbreviations feature from ParsedownExtra by using the following:
+
+```php
+use Erusev\ParsedownExtra\Features\Abbreviations;
+
+$State = Abbreviations::from(new State);
```
BenjaminHoegh, line 73: We have same issue here `new State` will by itself return a error so some info about how to handle it will help

aidantwoods: You're right, I should also be importing `State` next to the other `use` line :)

aidantwoods: This should be resolved in bf6cfe8

docs/Migrating-Extensions-v2.0.md
aidantwoods: I'm tempted to leave this example as is: just because there are three other classes here that also should really be imported too 😉  The intent of this example is really to give examples how to combine various "extensions", but not really meant to be run without importing all the mentioned classes.

aidantwoods:
> Really good so far, the only thing I can think of may confuse some people is when using extensions the example in this doc doesn't show you where you get `new State` from so by running the code example directly will return a error

Thanks for taking the time to review! I think I still need to add some more detail about adding custom blocks and inlines, and then also some explanation of the mechanics around parsing and interpreting objects like `Context`, `Line`, `Lines`, and `Excerpt`. And then also the different interfaces Blocks and Inlines can implement (e.g. `ContinuableBlock` or `BacktrackingInline`) that allow opting into some extra parsing features.

BenjaminHoegh:
A little question, while I was working on updating ParsedownMath to work with 2.0 I was hit by a question, how do you add a closing marker? As LaTeX doesn't use the same marker for both opening and closing.

BenjaminHoegh:
You are absolutely amazing

BenjaminHoegh:
I must say I have been enjoying updating parsedownMath for v2. When you first get how the new setup works, it is so easy to develop new functionality and so simple compared to v1. Great job I much say

aidantwoods:
> I must say I have been enjoying updating parsedownMath for v2. When you first get how the new setup works, it is so easy to develop new functionality and so simple compared to v1. Great job I much say

Thanks! That's really re-assuring feedback :)

Any parts that you feel it would be worth going over in extra detail, for when I finish these docs?

BenjaminHoegh:
> > I must say I have been enjoying updating parsedownMath for v2. When you first get how the new setup works, it is so easy to develop new functionality and so simple compared to v1. Great job I much say
> 
> Thanks! That's really re-assuring feedback :)
> 
> Any parts that you feel it would be worth going over in extra detail, for when I finish these docs?

Hmm good question, I think it may be the State object as this is a really important part

aidantwoods:
Thanks @BenjaminHoegh, I've added a little more of an explanation in [6be4366](https://github.com/erusev/parsedown/pull/821/commits/6be43664fd0e090a0e33fde891fff790659ad7b5), hopefully that is enough! 😄

Issue #746, opened 2019-12-31
backtick escaping inside a table seems to break the table


hartman:
Take the following markdown 

```markdown
| Formatting         | Symbol                   | Example                              |
| ------------------ | ------------------------ | ------------------------------------ |
| `Code`             | Three backticks \`\`\` | \`\`\`print 'Hello World';\`\`\`     |

```

In the first cell I want inline code markup, in the second cell, i want three escaped backticks, and in the third i want a pair of three escaped markdowns. As can be seen in the screenshot, made using https://parsedown.org/extra/ this breaks the table..
<img width="1696" alt="Screenshot 2019-07-11 at 14 03 51" src="https://user-images.githubusercontent.com/108913/61050127-8921a580-a3e6-11e9-9beb-a92fb4e2a6ef.png">

I couldn't find a way around it either.

Oh the irony of documenting a markdown flavour and finding an issue in a markdown parser which you then have to report using markdown escaped inside markdown ;)

PR #823, opened 2021-12-29
fix issue #746

MaximBobylev:
The problem described in the the issue could be simplified to parsing error in case of the following table
```
| header1 | header2 |
| --- | --- |
| ```a``` | ```b``` |
```
Instead of parsing values A and B into separate columns, these values are being concatinated into first column. PR adds one alternative into column value regex, which provides correct parsing of triple backtics

### DIFF
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -1033,7 +1033,7 @@ class Parsedown
             $row = trim($row);
             $row = trim($row, '|');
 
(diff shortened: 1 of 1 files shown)


k00ni:
Any update here? @erusev

BenjaminHoegh:
@erusev This issue is also present in the current state of the 2.0 version of Parsedown

PR #873, opened 2024-07-09
run tests using GitHub actions

xabbuh:


MyIgel:
Looks like this also reverted #871, was that intended behaviour @xabbuh / @erusev? That would be the only blocker for an intermediate release to fix PHP 8.4 compatibility as noted in #889

PR #876, opened 2024-07-17
add a branch alias to allow installing untagged 1.8 versions

xabbuh:


xabbuh:
@erusev Would be great to have this merged too as it would allow us to use these changes in the Twig tests without the need of having a new tag of the `1.8.x-beta` branch.

Issue #714, opened 2019-04-29
[Deprecation][v1.8.0-beta-7] List broken when used new line


laukstein:
Latest Parsedown beta version 1.8.0-beta-7, breaks list markdown containing new line like

    1. a
    
      b

turning to

    <ol>
    <li>
    <p>a</p>
    </li>
    </ol>
    <p>b</p>

While it works fine in latest stable v1.7.3, returning

    <ol>
    <li>
    <p>a</p>
    <p>b</p>
    </li>
    </ol>

aidantwoods:
The result given by 1.7.3 isn't correct—the line with "b" needs to be indented by one more space so that it matches up with the original list content's indentation.

laukstein:
@aidantwoods, you are wrong. I just rechecked it in https://parsedown.org/demo and I switched to Parsedown v1.7.3, and it works as I mentioned above.

aidantwoods:
> @aidantwoods, you are wrong. I just rechecked it in https://parsedown.org/demo and I switched to Parsedown v1.7.3, and it works as I mentioned above.

The result given ***by*** Parsedown 1.7.3 isn't correct (not *for*). I agree that Parsedown gives the result you mentioned, however it is this result which is incorrect.

The CommonMark reference parser indeed [agrees](https://spec.commonmark.org/dingus/?text=1.%20a%0A%0A%20%20b) with the new (1.8.x-beta) result, giving:
```html
<ol>
<li>a</li>
</ol>
<p>b</p>
```

To see why this is the case, see the spec description under [example 224](https://spec.commonmark.org/0.29/#example-224), which says:

> The most important thing to notice is that the position of the text after the list marker determines how much indentation is needed in subsequent blocks in the list item. If the list marker takes up two spaces, and there are three spaces between the list marker and the next non-whitespace character, then blocks must be indented five spaces in order to fall under the list item.

laukstein:
@aidantwoods, having space/s or not, 1.8.0-beta-7 currently fails on it and doesn't wrap inside <li>

aidantwoods:
As mentioned, it needs to be indented to **match** the indentation of the list content where the marker begins, like this:

<img width="565" alt="Screenshot 2019-04-29 at 20 37 56" src="https://user-images.githubusercontent.com/3288888/56921839-aff65e00-6abe-11e9-902e-f8b006fbb916.png">

laukstein:
[...]

PR #881, opened 2024-08-28
Merge `1.8.x-beta` into `master`

xPaw:
The changes are minimal, master already had all the code changes from 1.8 branch, so this simply adds github actions and bumps up the php version.

@erusev mind taking a look so that there's less confusion about master and the 1.8 branch?


erusev:
@xabbuh do you have any thoughts on this?

xabbuh:
@erusev sounds reasonable to me

erusev:
If we merge `1.8.x-beta` into `master`, can we then remove the `1.8.x-beta` branch? I find the branch situation a bit confusing.

xPaw:
I believe it can be removed, yes.

erusev:
> If we merge `1.8.x-beta` into `master`, can we then remove the `1.8.x-beta` branch? I find the branch situation a bit confusing.

@xabbuh do you have any thoughts on this?

aragon999:
One thing, which is changed by this merge in addition to the changes which are in the 1.8.x branch, is that the tests are not compatible with php 7.1 (phpunit version 7.5).

But in my opinion this is fine, as this version is already eol for some time now (December 2019): https://www.php.net/eol.php

Maybe this php version should be dropped as well for the 1.8.0 release.

xPaw:
PHPUnit 7 specifies PHP 7.1 - PHP 7.3 support, is that not the case?

The github action does have 7.1 in the test matrix, but it didn't run on this PR, so I'm not sure if it actually works.

aragon999:
@xPaw ah yes, sorry for the confusion. I was thinking about it the other way around, i.e. see the following commment:
https://github.com/sebastianbergmann/phpunit/blob/7.5.20/src/Framework/TestCase.php#L407

But if you add the `void` declaration in the child class, it is fine: https://3v4l.org/tNYS3#v7.1.33

xabbuh:
This looks reasonable to me. 👍

erusev:
Should we also release `1.8.0` and make it latest? @aidantwoods @xabbuh  @xPaw

xPaw:
I've been running that code in production for a while.

erusev:
What should the release notes for `1.8.0` say?

erusev:
@aidantwoods Would love to get your thoughts on this.

erusev:
Looks like initially @aidantwoods wanted to release this as 1.8:

> On making this a minor release, I think it is certainly "fair game" due to not breaking the public API. -- https://github.com/erusev/parsedown/issues/601#issuecomment-387080763

But then he decided against it:

> I ended up deciding not to release 1.8.x as a minor upgrade due to quite a few internal changes that might break extensions. -- https://github.com/erusev/parsedown/issues/714#issuecomment-487721882

I'm open to releasing it as `2.0`, but the fact that @aidantwoods never released it as a stable release makes me wonder if it's good enough.

joeworkman:
If you want to play it safe, release it as 2.0...

xabbuh:
@Evgeny1973  with which version of this package do you get it?

Evgeny1973:
@xabbuh Parsedown 1.7.4 and Symfony 7.1.

Evgeny1973:
Sorry, as I see fixes of these depressions to the master branch have already been made. Looking forward to the release!

dregad:
@xabbuh would it be possible to tag 1.7.5 with this fix ? see #889

PR #896, opened 2024-12-29
Update .gitattributes

kudashevs:
Hello,

This PR removes a redundant `.travis.yml` file from `.gitattributes`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,5 +2,4 @@
 /test               export-ignore
 /.gitattributes     export-ignore
 /.gitignore         export-ignore
-/.travis.yml        export-ignore
 /phpunit.xml.dist   export-ignore


.gitattributes
```diff
@@ -2,5 +2,4 @@
 /test               export-ignore
 /.gitattributes     export-ignore
 /.gitignore         export-ignore
-/.travis.yml        export-ignore
```
williamdes, line 5: Please add `.github` instead

.gitattributes
```diff
@@ -2,5 +2,4 @@
 /test               export-ignore
 /.gitattributes     export-ignore
 /.gitignore         export-ignore
```
williamdes, line 4:
```suggestion
/.gitignore         export-ignore
/.github            export-ignore
```

.gitattributes
kudashevs: Thanks for your review. I will keep my PR as is.

My intention is to delete a thing that no longer exist.
When the file is cleaned up, it can be improved further.

williamdes: You can click apply on my suggestion
that will make your pr more accurate

kudashevs: You can create your own PR if you want.

As I've said, my intention is different. Thank you.

williamdes: done in https://github.com/erusev/parsedown/pull/905

Issue #897, opened 2024-12-30
New release (1.8 / 2.0 ?)


dregad:
Dear maintainers,

Today marks the **5th anniversary** 🎂 of [Parsedown 1.7.4](https://github.com/erusev/parsedown/releases/tag/1.7.4)), which is the library's latest official release.

I was really happy to see some activity again this year, and would like to thank you all for your efforts !

Unfortunately the current state of the project is somewhat confusing at the moment... The 1.8.x branch has been merged into master (#881) in October, and the subsequent commit 999fcf788675b65b5344e2744443fc4d5f76cc14 bumped the version to 1.8.0, but there has not been any corresponding tag or release yet.

Mid-November, there's been some discussion about what the next version number should be (1.8.0 or 2.0.0), as well as concerns about the state of the code in the master branch (https://github.com/erusev/parsedown/pull/881#issuecomment-2472689241), but as far as I know that has stalled.

PHP 8.4 has been out for over a month now, and as you're probably aware there are deprecation warnings in the 1.7.x branch, which have been fixed in master (6ae01284b835bddc0264a2ac4e63f570b9457357) but can't be used with a regular composer update due to the lack of an official release.

It would be truly awesome if, to celebrate 2025, we could start off with a fresh new release... Please 🙏 

Thanks in advance and happy new year to everyone ! 🍾 🎉

bohwaz:
It seems that work on the 2.0 branch has stopped?

ArnaudLigny:
I propose a compromise: we abandon version 1.8 (because of the aborted betas) without going towards a major version (2.x) and... we jump to version 1.9.0 :-)

dregad:
Anything other than status quo would be good 😉

DigitalSorceress:
Came here because I had forked from 1.8.0-beta-7 at one point, and made a small setter change (to let me specify if it should set target attributes to _self or _blank

I admit I didn't do the best job with inline comments so I ended up diffing to see what I had changes - only to find that I don't even see the 1.8.0-beta-7 anywhere
and I was confused when I too saw that the 1.7.4 had .. included some stuff from the 1.8  betas

So yes, just wanted to say it would be nice to get that sorted.

For my part I think I'm going to go back to 1.7.4 latest and reapply my changes to that rather than risk having built off abandoned betas.

I've genuinely found parsedown to be a great little tool for letting me make web sites semi-static (the parsedown part is dynamic but all my content is in .md files ... making it low friction to write and publish new stuff.

[...]

PR #899, opened 2025-01-26
Add PHP 8.4 support for 1.7.x by dropping support for PHP < 7.1

MyIgel:
* Re-add optional nullable which got implemented in #871 and reverted in #873 (due to its PHP < 7.1 incompatibility?)
* Drop support for PHP 5.2, 5.4, 5.5, 5.6, 7.0 because only PHP 7.1 supports optional nullables (and is also unsupported for ages)
* Fix PHPUnit warning of missing name attribute in tests
* Fix CommonMark tests on PHP >= 7.2 by matching the inherited `setUp` return type


MyIgel:
This could also be the fix for #897 when released

Evgeny1973:
@erusev Is there going to be a new release anytime?

tacman:
Please don't sacrifice PHP 8.4 compatibility for long-dead versions of PHP or any other library.  Anyone still using something that old likely isn't updating this library, but is instead inside of code that hasn't be touched in years.

u01jmg3:
Could this be released as v2? Anyone updating their dependencies would explicitly have to update this dependency to be v2 thus preventing unwanted breakages.

xPaw:
To prevent confusion. All these changes are already merged in the master branch. This PR is specifically trying to update the 1.7.x branch.

tacman:
Great.  A new release would be welcome.

image

BafS:
You can use dev-master to have the latest version compatible with PHP8.4 (not the best practice but there is no other tags)

LukeWCS:
Hi

I'm currently updating a test suite for phpBB extensions, and I'm using this tool, among other things.

> Please don't sacrifice PHP 8.4 compatibility for long-dead versions of PHP or any other library.

I agree.

PHP 5 is no longer relevant today, and PHP 7 is also long outdated. You shouldn't implement "complicated" workarounds for outdated PHP versions, especially when the problem, like with implicitly nullable, is very compact and can be solved natively.

PR #905, opened 2025-02-17
Fix .gitattributes

williamdes:
This was missed in https://github.com/erusev/parsedown/pull/874

TravisCI config is now in the GitHub folder

Closes: https://github.com/erusev/parsedown/pull/896


dregad:
Fixed in https://github.com/parsedown/parsedown/pull/5

PR #915, opened 2025-05-06
Fix implicit null param deprecation warning in php 8.4

KEINOS:
This PR introduces a bare minimum updates for `Parsedown` 1.7.4 codebase (`erusev:1.7.x` branch) to avoid 2 major warnings. No other refactoring are made.

### Type Hinting Improvements:
* Updated the method signatures of `blockSetextHeader` and `blockTable` in `Parsedown.php` to allow `null` values for the `$Block` parameter by using nullable type hints (`?array`).
    * This avoids "[deprecation of implicit nullable parameter](https://wiki.php.net/rfc/deprecate-implicitly-nullable-types#code_change_examples)" warnings in PHP 8.4.

### PHPUnit Configuration Update:
* Added a descriptive name, "Parsedown Test Suite," to the `<testsuite>` element in the `phpunit.xml.dist` file.
    * This avoids "[config file validation error](https://docs.phpunit.de/en/9.6/configuration.html#the-testsuite-element)" warnings in PHPUnit 9.6.

### Add `.gitignore`:

* Added a `.gitignore` file to exclude "`vendor/`" dir and cache/lock files to git commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+vendor/
+composer.lock
+.phpunit.*.cache
+
--- a/Parsedown.php
+++ b/Parsedown.php
@@ -712,7 +712,7 @@ class Parsedown
     #
     # Setext
 
-    protected function blockSetextHeader($Line, array $Block = null)
+    protected function blockSetextHeader($Line, ?array $Block = null)
     {
         if ( ! isset($Block) or isset($Block['type']) or isset($Block['interrupted']))
         {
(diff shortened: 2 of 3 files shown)


KEINOS:
Ah ... now I notice that it's kind of "we've been there" stuff.

**This fix will break BC**, since the fix rely on "nullable type declarations with the `?T` syntax" from PHP 7.1. Thus, PHP 5.x, 6.x, 7.0 will not work with this fix 😭

PR #930, opened 2026-02-19
fix: Change constructor parameter to nullable type

### LINKED ISSUE #914: Deprecated: Erusev\Parsedown\Parsedown::__construct()
Deprecated: Erusev\Parsedown\Parsedown::__construct(): Implicitly marking parameter $StateBearer as nullable is deprecated, the explicit nullable type must be used instead in /var/www/html/vendor/erusev/parsedown/src/Parsedown.php on line 30

kumy:
Fix "Implicitly marking parameter $StateBearer as nullable is deprecated" since php 8.4

```
ErrorException: Deprecated: Erusev\Parsedown\Parsedown::__construct(): Implicitly marking parameter $StateBearer as nullable is deprecated, the explicit nullable type must be used instead
```

Closes #914

### DIFF
--- a/src/Components/Block.php
+++ b/src/Components/Block.php
@@ -17,6 +17,6 @@ interface Block extends Component
     public static function build(
         Context $Context,
         State $State,
-        Block $Block = null
+        ?Block $Block = null
     );
 }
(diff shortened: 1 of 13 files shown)


kumy:
@erusev Is there any other file that need to be updated to propose a `v2.0.0-beta-2`?

